PR #19, opened 2018-04-23
Some Tests! 🎉

### LINKED ISSUE #4: No Tests 🤦‍♂️

aviflax:
This introduces a bunch of specs, including some function specs (6, I think) and adds some tests that rely wholly on `stest.check` to perform property testing of those 6 functions. (I’m a big fan of property testing, AKA generative testing.)

Six functions isn’t fantastic coverage, in fact I haven’t even introduced coverage measurements yet, but this is really about laying the groundwork for greater coverage. Now that we have these specs and the plumbing in place for using them in tests, we should be able to greatly expand coverage without a crazy amount of work.

## Next Steps

After this PR merges:

* More tests!
* Start measuring coverage!
* Set up CI!

## Housekeeping

I plan to squash-merge this branch, but for now I’ve left all 18 commits as-is… it’s kinda interesting to see how this came together. I’m actually writing a blog post about the various approaches I took to creating the specs.

Fixes #4.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,4 +27,4 @@ COPY . ./
 # The max heap size is set to 1GB because Docker machines frequently have less than 4GB of RAM, and
 # JDK 8 defaults to setting the max heap to 1/4 of the total RAM, but such a small heap max leads
 # to OOM errors when running the tests (at 512MB; I’ve only tested 512MB and 1GB).
-ENTRYPOINT sleep 1; clj -J-Xmx1g -Atest
+ENTRYPOINT sleep 1; clj -J-Xmx1g -A:run-tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,15 +5,26 @@ LABEL maintainer="Avi Flax <avi.flax@fundingcircle.com>"
 WORKDIR /tmp
 RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
 RUN apk add --update --no-cache bash curl rlwrap@testing
-RUN curl -O https://download.clojure.org/install/linux-install-1.9.0.315.sh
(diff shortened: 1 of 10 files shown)


aviflax:
> This looks fine.

Great, thanks!

> CLJ tool usage checks out, although pending other build tool conversations I don't honestly see advantages to it.

I don’t want to dive too deep into a tangent here, so pending a more focused discussion I’ll just share that I _mainly_ used clj+tools.deps here in the spirit of experimentation. Now that I’ve got a little experience with them, I do see some potential advantages to sticking with them. But we can discuss those in a more suitable contex. (I’ll be in the SF office this week Tue–Thur so we have a somewhat rare opportunity to discuss in person if you’d like!)

> I really dislike this use of spec to exhaustively describe record types and their keys. See https://github.com/arrdem/detritus/blob/master/src/detritus/spec.clj for a macro I've been using to try and make this not as verbose and more reasonable.

Thanks for the food for thought! I read through that code and at first I couldn’t tell what it does, and I don’t have a REPL handy to try it out (I’m on a plane and forgot my laptop charger 🤦‍♂️). Reading through the docs for guten.tag did help me understand what it is you don’t like and your preferred alternative, at least WRT values if not specs. That said, it seems to me that this usage of spec and “bare” maps is quite idiomatic in the Clojure community, so I think I’d like to stick with it for now. That said, I’m going to keep thinking this over and I’ll try to try out that `spec` macro some time and see how it feels.

> I'm not sure what to do differently here, but the shrink spec kinda smells to me. Feels like it's trying to do too much, but I don't have enough context here to critique more carefully.

I don’t know; I’ve got the context, but when I read the implementation and spec for `shrink` they both seem within-bounds to me… at least, I don’t see an obvious way to simplify either. I’ll keep a lookout though.
 
> Other than that looks perfectly reasonable.

Great, thanks! And thanks for the review!

Issue #17, opened 2018-04-12
Bug in docstring for stop


aviflax:
As reported by @igoratron it references `start` which has since been renamed to `wcb`

PR #20, opened 2018-04-23
Some Docs! 🎉

### LINKED ISSUE #14: No Docs 🤦‍♂️
The repo has no docs right now! At the very least we should copy the docs from [The Toolset](https://fundingcircle.github.io/fc4-framework/methodology/4_Toolset.html).

Also try to explain the workflow!

Also: add a warning that comments will be lost!

/cc @igoratron

aviflax:
Non-zero amounts of docs!

Fixes #14.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,47 @@
 # FC4C
 
-A tool for reorganizing, restructuring, and reformatting [FC4](https://fundingcircle.github.io/fc4-framework/) diagrams.
+A tool for reorganizing, restructuring, and reformatting
+[FC4](https://fundingcircle.github.io/fc4-framework/) diagrams.
 
-# **WARNING: THIS IS A HACKY WIP.**
+## Purpose
 
-## Super-Sketchy How-To
+As explained in
+[The Toolset](https://fundingcircle.github.io/fc4-framework/methodology/toolset.html) section of
+[the FC4 Methodology](https://fundingcircle.github.io/fc4-framework/methodology/):
 
-1. Have `clj` installed ([guide](https://clojure.org/guides/getting_started))
-1. Run in your shell: `clj`
-1. Run in the REPL:
-   1. `(use 'fc4c.repl)`
-   1. Read the printed descriptions of `pcb` and `wcb`
-   1. Run either `(pcb)` or `(wcb)`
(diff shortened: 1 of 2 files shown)


aviflax:
Fixes #17

README.md
```diff
@@ -1,14 +1,56 @@
...
+4. To install the dependencies, run: `clj -e '(println "deps installed!")'`
+
+## Usage (Workflow)
+
+As explained in [The Authoring Workflow](https://fundingcircle.github.io/fc4-framework/methodology/authoring_workflow.html) section of
+[the FC4 Methodology](https://fundingcircle.github.io/fc4-framework/methodology/):
+
+> 1. In your text editor: either create a new diagram source file or open an existing diagram source file
+> 1. In a terminal, in your FC4C working dir:
+>    1. Open a Clojure REPL with `clj`
```
igoratron: ❓ forgot to ask last time, what is the benefit of using REPL here?

aviflax: Great question!

I absolutely intend for the tool to be runnable directly from the command line, and you can see there’s a little bit of such functionality there already in [`fc4c.core/-main`](https://github.com/FundingCircle/fc4c/blob/7d01df973294175316c4a0952bb877c8a8b328ca/src/fc4c/core.clj#L209). Right now all it does is process a YAML document via STDIN and output the result, which isn’t super useful, especially because of the startup time involved in starting up a Clojure tool wherein all the code, including the dependencies, needs to be compiled on each startup (it’s not AOT compiled).

So certainly I want to change this to be able to use the `wcb` (watch clipboard) workflow from the CLI.

I’ve only held off on doing so because I had been thinking that I also want to enable a `wfs` (watch filesystem) workflow via the CLI wherein an entire directory would be watched for changes and files would be processed in-place. And I wasn’t sure how to approach this — maybe with “commands” as args… probably, yeah, I just haven’t thought that through nor do I know how to implement this off the top of my head.

Now that you’ve asked this question, that seems silly. We have `wcb` now, we should make it easier to use. I’ll go ahead and do this now. So I’ve created #20.

Now that that’s set, I don’t think this needs to hold up _this_ PR. Agreed?

igoratron: sorry, this wasn't a blocker. Just asking out of curiosity.

aviflax: No worries! Thanks!

PR #27, opened 2018-05-23
CircleCI

aviflax:
This adds a CircleCI build config, and also:

* Switches the commands in the README and in the Dockerfile to use the `clojure` script rather than `clj`
  * As per [the tools.deps reference](https://clojure.org/reference/deps_and_cli#_usage), the `clojure` script is a runner for Clojure while `clj` is a wrapper for interactive repl use. 
* Bumps the max heap size to 2GB when running the tests from the Dockerfile
  * I saw a few OOM errors when testing with the max at 1GB
* Since I was already in the Dockerfile, I streamlined it by switching to a new base image that’s now available that already has Clojure and tools.deps installed.

[ENG-1481](https://jira.fundingcircle.com/browse/ENG-1481)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
    build:
      docker:
-       - image: circleci/clojure:tools-deps
+       - image: clojure:tools-deps-alpine
      steps:
        - checkout
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,11 @@ jobs:
        - run: "clojure -R:run-tests -e '(println \"deps are installed!\")'"
 
        # Run the tests
-       - run: "clojure -A:run-tests"
+       # The max heap size is set to 1GB because CI machines/containers frequently have <=4GB of
+       # RAM, and JDK 8 defaults to setting the max heap to 1/4 of the total RAM, but such a small
+       # heap max leads to OOM errors when running the tests (at 512MB; I’ve only tested 512MB and
+       # 1GB).
+       - run: "clojure -J-Xmx1g -A:run-tests"
 
        # Save the dependencies cache
        - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,10 +14,10 @@ jobs:
            - dependencies-
 
        # Download deps if necessary. (clj automatically downloads deps if necessary)
-       - run: clj -e '(println "deps are installed!")'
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
...
+       - run: "clojure -R:run-tests -e '(println \"deps are installed!\")'"
+
+       # Run the tests
+       # The max heap size is set to 2GB because I’ve seen OOM errors at 1GB and below. (JDK 8
+       # defaults to setting the max heap to ¼ of the total RAM, and CI containers frequently have
+       # <= 4GB RAM.)
+       - run: "clojure -J-Xmx2g -A:run-tests"
+
+       # Save the dependencies cache
+       - save_cache:
```
sgerrand: 🙋 Recommend saving the cache immediately after pulling down and installing the project dependencies. Not a blocker though.

.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
+version: 2
+jobs:
+   build:
+     docker:
+       - image: circleci/clojure:tools-deps
```
sgerrand: 🙋 Suggest using the same Docker image that is used as a base image in the `Dockerfile`.

.circleci/config.yml
aviflax: Ah yeah, that would be more logical. Thanks!

.circleci/config.yml
aviflax: That consistency would be good! The main reason I went with this is because I figured, given that it’s one of Circle’s official supported images, and it’s what their templates suggest, it might be more likely to be cached by their platform, which _could_ lead to faster builds. _If_ that’s real, it could be a tangible benefit, and could potentially outweigh the consistency benefit. Unless there’s some other potential benefit of using the same image in both places that I might be missing?

aviflax: Also, it occurs to me that the `checkout` step _might_ require that git be installed in the image… and I suspect it is not installed in `clojure:tools-deps`. I could test it though, if you think it’s worthwhile.

sgerrand: CircleCI have their own native client which gets used if a `git` binary isn't present in the main image. It should work fine using `clojure:tools-deps`.

aviflax: Aha, good to know. I’ll give it a try. Thanks!

aviflax: Tried it, it looks good. Thanks!

PR #31, opened 2018-06-19
New data structure/format representing a unified model of all systems and users

aviflax:
I know the PR stats make this look big, but it’s mostly test data in YAML files. In terms of code there’s “only” ~240 new lines.

Background from the (private within Funding Circle) ticket:

> Currently with FC4, we have only a single kind of data structure: a diagram. It defines both the systems and their elements and all the relationships thereof, and how those should all be depicted in a specific view.
> 
> This is very straightforward in the sense that we end up with a 1:1 correlation between YAML file and PNG file, and we have only one kind of file/structure: a diagram.
> 
> But! It has a bunch of downsides:
> 
> * Requires lots of duplication and repetition
>   * e.g. a container diagram is a superset of the same system’s context
>     diagram
>   * e.g. the same element may appear in multiple files, and its content
>     (data) (e.g. description) need to be repeated each time
> * That duplication and repetition leads to too much busywork to keep
>   things in sync, and too much risk that things will drift
> 
> Really, each thing should be defined/described/stated in one single time+place. A “single source of truth” — look at that, that idea pops up
> _everywhere!_
> 
> So let’s try embracing one of the aspects of
> [Structurizr](https://structurizr.com/) that I had initially dismissed: defining the _model_ of one’s systems independently of the _views_ of those systems.
> 
> Concretely, this would mean that instead of a single dir tree containing simply _diagrams_, we’d have a bifurcated tree containing the _model_ and the _views_ (or perhaps we should simply call the views _diagrams_).

This PR specifies the data structure and YAML file-format for capturing the model of a system landscape: a set of interrelated systems and users. It also includes functions for reading the model from those YAML files on disk — in a specific directory structure — and validating the model once it’s been read.

(Ideally this would include changes to the docs, to describe the new directory structure, file formats, etc. I’m holding off on that for now though, because those changes would make the most sense to be applied to the “methodology” documentation in the fc4-framework repo. I recently concluded that these two repos need to be unified; I plan to move the tool(s) herein to a new dir named `tools` in the framework repo, unifying the framework with its docs and tools in a single repo. This will also resolve the question of this tool(set) needing a new name: it’ll just be _FC4 Tools_ or in code `fc4.tools.*`)

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -2,29 +2,37 @@
  {org.clojure/core.async {:mvn/version "0.4.474"}
   circleci/clj-yaml {:mvn/version "0.5.6"}
   potemkin {:mvn/version "0.4.4"}
+  expound {:mvn/version "0.7.0"}
+  com.cognitect/anomalies {:mvn/version "0.1.12"}
 
-  ;; test.chuck is in the main dependency list, rather than in the test profile, because we’re
-  ;; using one of its generators in a spec in a source file.
+  ;; Normally test.check would only be present in profiles like :dev, :test,
+  ;; etc. That way it wouldn’t need to be included in distributed/built
+  ;; artifacts nor downloaded/used at runtime. Unfortunately we have a few
+  ;; generators that need to use let, which for some reason isn’t included in
+  ;; clojure.spec.gen.alpha. We might be able to make our own lazy-loading alias
+  ;; for let. So TBD I guess.
+  org.clojure/test.check {:mvn/version "0.9.0"}
+
+  ;; test.chuck is in the main dependency list, rather than in the test profile,
+  ;; because we’re using one of its generators in a spec in a source file.
   com.gfredericks/test.chuck {:mvn/version "0.2.8"}}
 
  :aliases {:dev       {:extra-paths ["test"]
-                       :extra-deps  {org.clojure/test.check     {:mvn/version "0.9.0"}
+                       :extra-deps  {org.clojure/tools.trace    {:mvn/version "0.7.9"}
+                                     inspectable                {:mvn/version "0.2.2"}
                                      com.cognitect/test-runner  {:git/url "https://github.com/cognitect-labs/test-runner"
                                                                  :sha "76568540e7f40268ad2b646110f237a60295fa3c"}}}
            :test      {:extra-paths ["test"]
-                       :extra-deps  {org.clojure/test.check     {:mvn/version "0.9.0"}
-                                     com.cognitect/test-runner  {:git/url "https://github.com/cognitect-labs/test-runner"
+                       :extra-deps  {com.cognitect/test-runner  {:git/url "https://github.com/cognitect-labs/test-runner"
                                                                  :sha "76568540e7f40268ad2b646110f237a60295fa3c"}}}
            :run-tests {:extra-paths ["test"]
-                       :extra-deps  {org.clojure/test.check     {:mvn/version "0.9.0"}
-                                     ;; This fork returns a non-zero status when tests fail; crucial when building in a CI server.
+                       :extra-deps  {;; This fork returns a non-zero status when tests fail; crucial when building in a CI server.
                                      ;; See https://github.com/cognitect-labs/test-runner/pull/12
                                      com.cognitect/test-runner  {:git/url "https://github.com/Olical/test-runner"
                                                                  :sha "7c4f5bd4987ec514889c7cd7e3d13f4ef95f256b"}}
                        :main-opts   ["-m" "cognitect.test-runner"]}
            :coverage {:extra-paths  ["test"]
-                      :extra-deps   {org.clojure/test.check     {:mvn/version "0.9.0"}
-                                     cloverage                  {:mvn/version "1.0.10"}}
+                      :extra-deps   {cloverage                  {:mvn/version "1.0.10"}}
                       :main-opts    ["-m" "cloverage.coverage"
                                      ;; TODO: make this regex exclude fc4c.test-utils
                                      "--ns-regex" "^fc4c.+(?<!test)"
--- a/deps.edn
+++ b/deps.edn
@@ -31,13 +31,18 @@
                                      com.cognitect/test-runner  {:git/url "https://github.com/Olical/test-runner"
                                                                  :sha "7c4f5bd4987ec514889c7cd7e3d13f4ef95f256b"}}
                        :main-opts   ["-m" "cognitect.test-runner"]}
-           :coverage {:extra-paths  ["test"]
-                      :extra-deps   {cloverage                  {:mvn/version "1.0.10"}}
-                      :main-opts    ["-m" "cloverage.coverage"
-                                     ;; TODO: make this regex exclude fc4c.test-utils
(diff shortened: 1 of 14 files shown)


deps.edn
```diff
@@ -2,34 +2,47 @@
...
+                       :extra-deps   {cloverage                  {:mvn/version "1.0.10"}}
+                       :main-opts    ["-m" "cloverage.coverage"
+                                      ;; TODO: make this regex exclude fc4c.test-utils
+                                      "--ns-regex" "^fc4c.+(?<!test)"
+                                      "--test-ns-regex" "fc4c.+\\-test$"
+                                      "--fail-threshold" "65"
+                                      "--codecov"]}
+           :lint      {:extra-deps {com.jameslaverack/cljfmt-runner
+                                    {:git/url "https://github.com/JamesLaverack/cljfmt-runner"
+                                     :sha "51f85c9d6cc67107620749ceacd46120647fe107"}}
```
fussybeaver: Do you need to use this particular version of the linter? If so, doesn't this need to also go in the `fix` command below ?

deps.edn
```diff
@@ -2,34 +2,47 @@
...
-                                     com.cognitect/test-runner  {:git/url "https://github.com/cognitect-labs/test-runner"
+                       :extra-deps  {com.cognitect/test-runner  {:git/url "https://github.com/cognitect-labs/test-runner"
                                                                  :sha "76568540e7f40268ad2b646110f237a60295fa3c"}}}
            :run-tests {:extra-paths ["test"]
-                       :extra-deps  {org.clojure/test.check     {:mvn/version "0.9.0"}
-                                     ;; This fork returns a non-zero status when tests fail; crucial when building in a CI server.
+                       :extra-deps  {;; This fork returns a non-zero status when tests fail; crucial when building in a CI server.
                                      ;; See https://github.com/cognitect-labs/test-runner/pull/12
                                      com.cognitect/test-runner  {:git/url "https://github.com/Olical/test-runner"
                                                                  :sha "7c4f5bd4987ec514889c7cd7e3d13f4ef95f256b"}}
```
fussybeaver: I think this is using a different version of the test runner between `test` and `run-tests`.. which should probably not be.

src/fc4c/model.clj
```diff
@@ -0,0 +1,323 @@
...
+       (str (->> (repeat 5 s) (join "/")) "/"))))
+
+(s/def ::file
+  (partial instance? java.io.File))
+
+(s/def ::dir-path-or-file
+  (s/or ::dir-path ::file))
+
+(defn- get-tags-from-path
+  [file relative-root]
```
fussybeaver: Would be good to have a docstring for this fn, as I'm slightly confused on how `::tags` work.

src/fc4c/model.clj
```diff
@@ -0,0 +1,323 @@
...
+        :args (s/cat :tags-from-path ::tags
+                     :proto-element  ::proto-element)
+        :ret ::element)
+
+;; A file might contain a single element (as a map), or an array containing
+;; multiple elements.
+(defn- elements-from-file
+  "Parses the contents of a YAML file, then processes those contents such that
+  each element conforms to ::element. The file-path and root-path are used to
+  generate tags from the file’s path relative to the root path."
```
fussybeaver: :trophy:

src/fc4c/model.clj
```diff
@@ -0,0 +1,323 @@
...
+
+(s/fdef elements-from-file
+        :args (s/cat :file-contents ::yaml-file-contents
+                     :file-path     ::dir-path
+                     :root-path     ::dir-path)
+        :ret  (s/coll-of ::element))
+
+;;;; The below functions do I/O. The function specs are provided as a form of
+;;;; documentation and for instrumentation during development. They should not
+;;;; be used for generative testing.
```
fussybeaver: hmm... maybe these functions should be in another namespace?

deps.edn
aviflax: > Do you need to use this particular version of the linter?

Well, this new dependency resolution tool ([tools.deps](https://github.com/clojure/tools.deps.alpha)) is explicitly and intentionally a little different than most tools in the category. While it supports resolving and downloading artifacts (such as JAR files) from repositories (such as Maven) this is largely for backwards compatibility. The tool really wants people to take advantage of the fact that git is a [content addressable storage](https://en.wikipedia.org/wiki/Content-addressable_storage) (CAS) tool, and now that git is widespread, it’s possible to specify a dependency with extreme precision, using a version identifier that is inextricably tied to the actual code. This way one knows _exactly_ what one will get, and it can’t change.

(This has been presented by Alex Miller in [Dependency Heaven](https://www.youtube.com/watch?v=sStlTye-Kjk) and by Rich Hickey in [Spec-ulation](https://www.youtube.com/watch?v=oyLBGkS5ICk) (which is not about clojure.spec; RH has a history of light trolling with his talk titles).)

Anyway, sorry for the wall of text, but to wrap up: I _think_ I’m using this tool (tools.deps) the way it’s meant to be used, i.e. going with the grain of the tool. In the case of this particular library, no artifact has been published to Maven, so the only way to reference it is by its GitHub repo URL. And while tools.deps might referencing tags rather than commit hashes, it’s irrelevant in this case because this library’s repo contains no tags.

Sorry if that was TMI. Let me know!

> If so, doesn't this need to also go in the fix command below?

No, because these profiles are composable at runtime.

I’ll add this to the docs, but basically the profiles can be combined when invoking `clojure` like so: `clojure -A:lint:lint/fix`

I didn’t know this myself until I learned it from the readme of this lint-tool-runner library; now that I know it, I might be able to remove some duplication from other profiles in this file. I’ll try to take a look at some point.

deps.edn
aviflax: Good catch — no, it should not! I’ll post a follow-up PR to fix this.  Thanks!

aviflax: Ah, actually, it should. As per the comment, this dep is a _fork_ of `https://github.com/cognitect-labs/test-runner` — even though the name/key is the same. I think that was an error on my part; it’s a little confusing, maybe even Wrong. I’ll fix this in a follow-up PR. Thanks!

src/fc4c/model.clj
aviflax: Agreed. I’ll do this tomorrow. Thanks!

deps.edn
fussybeaver: Sounds fine if you need it. 👍 Just wondering, because the profiles in this project are a little more involved than usual.

aviflax: Yeah, I agree. Your comment prompted me to take a closer look at the profiles, and they can definitely be simplified. I’ve already done so in a local branch and I’ll post it up as a PR after I’ve merged this one.

src/fc4c/model.clj
aviflax: 🤔 hmm, maybe. Like, what, `fc4c.model.io/read` ? It’s a bit cumbersome, but fairly appealing; I do like the idea of keeping functions that do I/O cordoned off somewhere (“at the edges”). I’ll think it over. Thanks!

aviflax: I’m gonna try this in a subsequent branch.

PR #34, opened 2018-06-25
 Switch test runner to eftest for faster tests and builds

aviflax:
Our tests are computationally bound and the prior test runner was
running them sequentially. This is silly, since most machines used by
developers and CI are multi-core. Since [eftest](https://github.com/weavejester/eftest) runs tests across
multiple threads by default, I thought it might work well to address
this — and it does! On my machine, wall-clock time to run the tests
drops from ~66 seconds to ~32 seconds.

I had to upgrade test.check from the latest stable release, 0.9.0, to
the latest alpha release, 0.10.0-alpha3, as there’s some incompatibility
between 0.9.0 and eftest. See https://github.com/weavejester/eftest/issues/32

As long as I was writing custom test running code, I thought I may as
well have it product a JUnit XML file as well, so CircleCI can analyze
and render test stats. Turns out eftest made this real easy. Nice.

Having our own custom test runner should also make it easy to integrate
the measurement of test/code coverage into the test runs.

This eftest ticket should point the way towards that:
  https://github.com/weavejester/eftest/issues/50

Even if the requested feature isn’t implemented in eftest, the linked
commit in the CircleCI test runner (circleci.test) may show how we can
do the same.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,4 +15,4 @@ COPY . ./
 
 # The max heap size is set to 2GB because I’ve seen OOM errors at 1GB and below. (JDK 8 defaults to
 # setting the max heap to ¼ of the total RAM, and containers frequently have <= 4GB RAM.)
-ENTRYPOINT clojure -J-Xmx2g -A:run-tests
+ENTRYPOINT clojure -J-Xmx2g -A:test:test/run
--- a/src/fc4c/test.clj
+++ b/src/fc4c/test.clj
@@ -10,6 +10,7 @@
 ;;    see: https://github.com/weavejester/eftest/issues/50
 
 (def test-dir "test")
+
 (def output-path
   "This is optimized for CircleCI: https://circleci.com/docs/2.0/configuration-reference/#store_test_results"
   "target/test-results/eftest/results.xml")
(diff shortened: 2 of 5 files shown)


src/fc4c/test.clj
```diff
@@ -0,0 +1,46 @@
...
+     ;; I tested the other supported value (:namespaces) and it was
+     ;; significantly slower.
+     :multithread? :vars}))
+
+(defn -main []
+  (let [tests (find-tests test-dir)
+        results (run-tests tests opts)
+        exit-code (->> (select-keys results [:fail :error])
+                       vals
+                       (reduce +))]
```
fussybeaver: Maybe cleaner to fix the exit code to 1 or use something between 2 and 126, because there are POSIX standards beyond that, and an upper limit within the unsigned short range.

Dockerfile
```diff
@@ -15,4 +15,4 @@ COPY . ./
 
 # The max heap size is set to 2GB because I’ve seen OOM errors at 1GB and below. (JDK 8 defaults to
 # setting the max heap to ¼ of the total RAM, and containers frequently have <= 4GB RAM.)
-ENTRYPOINT clojure -J-Xmx2g -A:run-tests
+ENTRYPOINT clojure -J-Xmx2g -A:test:test/run
```
fussybeaver: Shouldn't the entrypoint be to run the `wcb` tool as described in the README ? This looks like it runs the tests.

aviflax: The purpose of this Dockerfile is to support development and testing. It wouldn’t work for `wcb` as `wcb` stands for Watch ClipBoard — the workflow depends entirely on reading from and writing to the clipboard — and a Docker container can’t access the host system clipboard.

Clearly, the purpose of this Dockerfile is unclear… I could add some comments but I wonder if I should just delete it. Maybe violates YAGNI.

PR #37, opened 2018-06-26
New data structure + file format for FC4 Views

aviflax:
Building on 97e611c78 this introduces a new data structure and file
format for defining System Context diagrams and Container diagrams that
can be exported to Structurizr Express (SE) format. After this we just
need a data struct + file format for styles, and then we’ll be able to
move on to exporting these views to SE diagrams.


aviflax:
Wow,  thanks for the rapid and effusive review! 🙇

PR #41, opened 2018-06-27
Minor test runner runner improvements

aviflax:
Please see commits.


sgerrand:
My 💭s on this are that if the test runner is in `src/` then it's reasonable to write tests for it. Perhaps you should move it somewhere else instead.

aviflax:
@sgerrand I like it! I’ll move it somewhere in `test`. Thanks!

PR #42, opened 2018-06-27
Add sad-path tests for io ns + fix bugs revealed

aviflax:
Happy to report that adding these tests revealed some bugs, and I fixed them.


sgerrand:
🤔 Perhaps `test/data/` needs to be ignored?

aviflax:
Why?

I mean, that’s the test data. Why wouldn’t we want that in the repo?

(Longer term, I want to move this all into [the fc4-framework repo](https://github.com/FundingCircle/fc4-framework). The code and the invalid test data will go under a `tools` directory; the _valid_ test data will also serve as examples used by and embedded in the (methodology) docs. But that’s at least a few weeks off. For now it seems reasonable to me to have this data right here.)

sgerrand:
>> 🤔 Perhaps test/data/ needs to be ignored?
>
> Why?
>
> I mean, that’s the test data. Why wouldn’t we want that in the repo?

Apologies, I should been explicit about what I meant by "ignoring". I was referring to code coverage.

aviflax:
Aha, no worries — that makes sense! There are a few files being measured that shouldn’t be, I’m definitely going to fix that, but I’ll do so in a follow-up PR.

PR #47, opened 2018-06-29
Feature: export FC4 views to Structurizr Express diagrams

aviflax:
System Context diagrams, specifically. More diagram types coming soon.

#### Before Merging:

* [x] Repoint to `master` (after #46 has been merged)

### DIFF
--- a/src/fc4c/integrations/structurizr/express/export.clj
+++ b/src/fc4c/integrations/structurizr/express/export.clj
@@ -1,0 +1,251 @@
+(ns fc4c.integrations.structurizr.express.export
+  "Functions concerned with exporting FC4 views as Structurizr Express diagrams."
+  (:require [cognitect.anomalies  :as anom]
+            [clj-yaml.core        :as yaml]
+            [clojure.walk         :as walk :refer [postwalk]]
+            [clojure.spec.alpha   :as s]
+            [clojure.string       :as string :refer [includes? join]]
+            [fc4c.integrations.structurizr.express.spec] ;; for side fx
+            [fc4c.io              :as io]
+            [fc4c.model           :as m]
+            [fc4c.styles          :as st]
+            [fc4c.util            :as u :refer [update-all]]
+            [fc4c.view            :as v]
+            [fc4c.yaml            :as fy :refer [split-file]]))
+
+(defn- position
+  [sys-name view]
+  (get-in view  (if (= sys-name (::v/system view))
+                  [::v/positions ::v/subject]
+                  [::v/positions ::v/other-systems sys-name])))
+
+(defn- maybe-add-in-house-tag
+  [tags]
+  (if (contains? tags :external)
+    tags
+    (conj tags :in-house)))
+
+(s/fdef maybe-add-in-house-tag
+        :args (s/cat :tags ::m/tags)
+        :ret  ::m/tags)
+
+(defn- replace-internal-tag
+  "The tag “internal” is a special reserved tag for Structurizr Express; for
+  diagrams that indicate an “inside” and an “outside”, such as System Landscape
+  diagrams and Container diagrams, the set of elements that have “internal”
+  (case insensitive) is used to draw the boundary box. So when exporting an FC4
+  view+model+styles as Structurizr Express diagram, we need to transform the
+  tag :interal to something that is _not_ reserved, so we use :in-house."
+  [tags]
+  (if-not (contains? tags :internal)
+    tags
+    (-> tags
+        (disj :internal)
+        (conj :in-house))))
+
+(s/fdef replace-internal-tag
+        :args (s/cat :tags ::m/tags)
+        :ret  ::m/tags)
+
+(defn- tags
+  [elem]
+  (->> (::m/tags elem)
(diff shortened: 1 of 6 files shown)


src/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -0,0 +1,251 @@
...
+        users (users-of subject-name model)]
+    (-> (map #(dep->relationship % subject-name) deps)
+        (concat (mapcat #(user->relationships % subject-name) users))
+        (add-control-points view)
+        ;; In the model, two different containers of the same system can have
+        ;; relationships to the same *other* system. In a container diagram
+        ;; those relationships may be meaningfully different. In a System
+        ;; Context diagram, however, those relationships are essentially
+        ;; duplicates and can and should be collapsed down to a single
+        ;; relationship.
```
fussybeaver: would it be useful to name this function "dedup-relationships" or "unique-system-relationships" to clarify this point ?

src/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -0,0 +1,251 @@
...
+    :styles (rename-internal-tag styles)
+    :size (::v/size view)}))
+
+(s/fdef view->system-context
+        :args (s/cat :view ::v/view
+                     :model ::m/model
+                     :styles ::st/styles)
+        :ret :structurizr/diagram)
+
+;; TEMP TEMP
```
fussybeaver: I'm guessing this was used as a REPL sandbox - and is not needed for the actual commit ?

src/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -0,0 +1,251 @@
...
+  "Add control points to relationships, when they’re specified in the view."
+  [rels
+   {subject-name ::v/system
+    {point-groups ::v/system-context} ::v/control-points
+    :as view}]
+  (->> rels
+       (group-by (partial relationship-with subject-name))
+       (mapcat (fn [[other-side-name these-rels]]
+                 (->> (get point-groups other-side-name)
+                      (inject-control-points these-rels))))))
```
fussybeaver: Would be good to have some more granular unit test cases for the more complex transformations in this namespace like `inject-control-points`, `add-control-points` and `users-of`.

src/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -0,0 +1,251 @@
...
+        :args (s/cat :tags ::m/tags)
+        :ret  ::m/tags)
+
+(defn- replace-internal-tag
+  "The tag “internal” is a special reserved tag for Structurizr Express; for
+  diagrams that indicate an “inside” and an “outside”, such as System Landscape
+  diagrams and Container diagrams, the set of elements that have “internal”
+  (case insensitive) is used to draw the boundary box. So when exporting an FC4
+  view+model+styles as Structurizr Express diagram, we need to transform the
+  tag :interal to something that is _not_ reserved, so we use :in-house."
```
fussybeaver: :trophy:

src/fc4c/integrations/structurizr/express/export.clj
aviflax: Yup. This is removed in #48.

src/fc4c/integrations/structurizr/express/export.clj
aviflax: Agreed! I’ll get on that. Thanks!

src/fc4c/integrations/structurizr/express/export.clj
aviflax: Hmm. The fact that the relationships are deduplicated is not the fundamental purpose of the function, which is to simply create the values for the relationships array in the exported SE diagram. I suppose it could be something like `derive-relationships` or `make-relationships` but I thought maybe something a bit simpler and perhaps a bit more declarative would be OK.

Also: this currently supports only diagrams of the _System Context_ type. Soon I’ll need to add support for _System Landscape_ and _Container_ diagrams. That will probably require significant changes to this function, or perhaps multiple variants of it. And/ perhaps a deduplicating function will be passed in to it… not sure. So I’m inclined to leave this as-is for now and reassess when adding that support for additional diagram types, which should be soon.

How’s that sound?

src/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -0,0 +1,251 @@
...
+            [fc4c.view            :as v]
+            [fc4c.yaml            :as fy :refer [split-file]]))
+
+(defn- position
+  [sys-name view]
+  (get-in view  (if (= sys-name (::v/system view))
+                  [::v/positions ::v/subject]
+                  [::v/positions ::v/other-systems sys-name])))
+
+(defn- maybe-add-in-house-tag
```
aviflax: Note to self: let’s remove the prefix `maybe-` from this function name; it’s inconsistent with `replace-internal-tag` below, and this degree of explicitness is unnecessary.

aviflax:
Thanks for the review @fussybeaver!

PR #48, opened 2018-06-29
New CLI command `export`

aviflax:
#### Before Merging:

* [x] Repoint to `master` (after #47 has been merged)


export
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
fussybeaver: maybe it's time to have a `bin` folder, since there are now several executables in the root folder ?

aviflax: Good idea! I’m planning to add one or two more commands soon (`lint` and maybe something like `edit` or `layout`) so I’ll do that then. Thanks!

PR #51, opened 2018-07-13
Tests++

aviflax:
When @fussybeaver reviewed #47 he suggested adding a few tests for a few functions would make those functions more readable, understandable, and maintainable. I thought it’d be trivial, so I got started…

In the message for 5462d3b I explained what happened next:

> Starting with b319dd77e and then in some subsequent commits, I added a bunch of :fn property tests to a bunch of function specs in
> ..express.export. I did so casually at first; the export feature seemed to be working just fine so I figured these tests were just pro forma and would just pass right off the bat.
> 
> But then... turns out, generative testing can unearth a ton of assumptions one made when implementing a function and writing its specs, by discovering edge cases that don’t behave quite as one imagined. This then forces one to consider what *should* a given function actually do, come to think of it, when faced with a certain edge case? What input should be considered valid, and what’s invalid? How should invalid input be handled? Should errors be thrown? Etc. (I decided to try to follow Postel’s law as closely as possible.)
> 
> This was a somewhat frustrating but ultimately enlightening experience; I think next time I start adding specs and property tests I will be a little more prepared and have an idea of what to expect — and therefore I’ll probably do so earlier in the process of implementing a feature.

So… this patch is just slightly bigger than I had anticipated… 😅 honestly it’s way too big. I’m not sure how I could break it up into a smaller chunk though. Reviewing each commit individually might work, but this might be something we need to get together and review together, live.

### DIFF
--- a/src/main/fc4c/integrations/structurizr/express/export.clj
+++ b/src/main/fc4c/integrations/structurizr/express/export.clj
@@ -10,15 +10,23 @@
             [fc4c.model           :as m]
             [fc4c.spec            :as fs]
             [fc4c.styles          :as st]
-            [fc4c.util            :as u :refer [update-all]]
+            [fc4c.util            :as fu :refer [update-all]]
             [fc4c.view            :as v]
             [fc4c.yaml            :as fy :refer [split-file]]))
 
+(fu/ns-with-alias 'structurizr 'sz)
+
 (defn- position
+  "Returns the position of the named system in the view."
   [sys-name view]
   (get-in view  (if (= sys-name (::v/system view))
                   [::v/positions ::v/subject]
                   [::v/positions ::v/other-systems sys-name])))
+
+(s/fdef position
+        :args (s/cat :sys-name ::m/name
+                     :view     ::v/view)
+        :ret  ::v/coord-string)
 
 (defn- add-in-house-tag
   [tags]
@@ -28,7 +36,12 @@
 
 (s/fdef add-in-house-tag
         :args (s/cat :tags ::m/tags)
-        :ret  ::m/tags)
+        :ret  ::m/tags
+        :fn   (fn [{{in-tags :tags} :args, out-tags :ret}]
+                (if (:external in-tags)
+                  (= in-tags out-tags)
+                  (and (contains? out-tags :in-house)
+                       (= in-tags (disj out-tags :in-house))))))
 
 (defn- replace-internal-tag
   "The tag “internal” is a special reserved tag for Structurizr Express; for
@@ -46,7 +59,13 @@
 
 (s/fdef replace-internal-tag
         :args (s/cat :tags ::m/tags)
-        :ret  ::m/tags)
+        :ret  ::m/tags
+        :fn   (fn [{{in-tags :tags} :args, out-tags :ret}]
+                (if (:internal in-tags)
+                  (and (contains? out-tags :in-house)
+                       (not (contains? out-tags :internal))
+                       (= (count in-tags) (count out-tags)))
+                  (= in-tags out-tags))))
 
 (defn- tags
   [elem]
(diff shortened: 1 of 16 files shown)


aviflax:
OK, here’s one idea — if reviewers don’t want to review each commit individually, I guess I could break this into a few PRs like so:

* Preparatory refactoring, such as improvements to `fc4c.test-utils`, could be one or more standalone PRs
* Changes to the model files (model.clj, model_specs.clj, model_test.clj)
* Changes to the view files
* Changes to the export files

The only significant downside I can think of would be that I’d have to squash all the above commits into 1, then split it again… losing the above commit history. Personally I like having a commit history that reflects and preserves how the work was actually accomplished.

src/main/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -149,49 +188,68 @@
...
-  (->> rels
-       (map-indexed (fn [i rel]
-                      (if-let [points (nth point-groups i nil)]
-                        (assoc rel :vertices points)
-                        rel)))))
+  ;; You might wonder: why use map-indexed and nth instead of just mapping over
+  ;; both collections with plain old map? It’s because plain old map will just
+  ;; skip/drop elements if the colls are of different sizes. In this scenario
+  ;; there may be fewered point-groups than relationships, as not all
+  ;; relationships will have point-groups. Hence this convoluted approach.
```
Mongey: Does this also assumes that the `(size rels)`> `(size point-groups)`

src/main/fc4c/integrations/structurizr/express/yaml.clj
```diff
@@ -2,11 +2,20 @@
   (:require [clojure.string :as str]
             [fc4c.yaml :as fy]))
 
+(defn- wrap-coord-strings
+  "If an entire value looks like a coordinate, wrap it in single quotes so as to
+  force it to a string, because otherwise it might be parsed by Structurizr
+  Express as a number, and therefore an invalid coordinate. e.g. `82,34` is a
+  valid number in European locales, and some YAML parsers will therefore parse
```
Mongey: 😵 oh yaml

src/main/fc4c/model_specs.clj
```diff
@@ -30,7 +30,18 @@
              :gen-max 10))
 
 (s/def ::repos ::small-set-of-keywords)
-(s/def ::tags ::small-set-of-keywords)
+
+(s/def ::tags
+  (s/with-gen ::small-set-of-keywords
+    #(gen/set (gen/one-of [(s/gen ::short-simple-keyword)
```
Mongey: 💭  Interesting, I guess this relates to my questions / confusion above about spec.

src/main/fc4c/integrations/structurizr/express/export.clj
```diff
@@ -144,27 +167,34 @@
...
             (= (:name ret) user-name)
 
             :user-not-in-model
-            (nil? ret))))
+            (and (includes? (:name ret) user-name)
+                 (includes? (:name ret) "undefined")))))
 
 (defn- deps-of
   "Returns the systems that the subject system uses — its dependencies."
+  ;; TODO: I think maybe we also need the
```
Mongey: 📝 something missing? `... also need the ... `

src/main/fc4c/model_specs.clj
aviflax: yup!

src/main/fc4c/integrations/structurizr/express/yaml.clj
aviflax: Yup

aviflax:
Thanks for the review Conor! I really appreciate it. I’ll reply to your specific comments soon. I’m interested to hear your thoughts on Postel’s law — I’m a fan.

Oh and I’m also going to look into the coverage situation before merging — it was looking good before that last commit, which was really just some readability improvements — TBD.

src/main/fc4c/integrations/structurizr/express/export.clj
aviflax: Well, `>=` actually — good point. Do you think there should be some assertion of that in the spec and/or the function itself?

src/main/fc4c/integrations/structurizr/express/yaml.clj
aviflax: When I spoke at clojure/nyc on Wednesday, I quipped that YAML is very easy for humans to read… and proportionally difficult for computers to read.

aviflax:
I’ve figured out how to fix the coverage. I’ll push one more commit either today or Monday that should fix it.

aviflax:
OK, I pushed 4 new commits, starting with f368a76, that fix the coverage, more or less. @Mongey would you mind reviewing those 4 new commits? Thanks!

PR #56, opened 2018-08-01
Bug fixes

aviflax:
Biggest bug fixed herein is that the snap-to-virtual-grid feature wasn’t working at all. But I found a few other bugs while looking into that.

It’d probably be best/easiest to review this commit-by-commit.

### DIFF
--- a/src/main/fc4c/integrations/structurizr/express/edit.clj
+++ b/src/main/fc4c/integrations/structurizr/express/edit.clj
@@ -129,11 +129,11 @@
                   (and
                     ; Yeah, OrderedMaps are equal to maps with the same entries
                     ; regardless of order — surprised me too!
-                    (= m ret)
-                    (= (or (keys ret) []) ; keys on empty map returns nil
-                       (concat (filter (partial contains? m) ks)
-                               (sort (remove (partial contains? kss)
-                                             (keys m)))))))))
+                   (= m ret)
+                   (= (or (keys ret) []) ; keys on empty map returns nil
+                      (concat (filter (partial contains? m) ks)
+                              (sort (remove (partial contains? kss)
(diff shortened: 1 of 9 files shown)


src/main/fc4c/integrations/structurizr/express/edit.clj
```diff
@@ -184,19 +212,28 @@
 (defn snap-coords
   "Accepts a seq of X and Y numbers, and config values and returns a string in
   the form \"x,y\"."
+  ; TODO: it’s inconsistent that the min-margin is passed in as an arg but the
```
sgerrand: 🙋 Please migrate this comment to an issue instead.

src/main/fc4c/integrations/structurizr/express/spec.clj
```diff
@@ -9,7 +9,10 @@
 
 (def ^:private aliases-to-namespaces
   {'st 'structurizr
+   'sy 'structurizr.system
+   'sp 'structurizr.person
    'sc 'structurizr.container
+   'sm 'structurizr.component
```
sgerrand: 🙋 Suggest sorting this list.

src/main/fc4c/integrations/structurizr/express/edit.clj
aviflax: Good idea, will do — thanks!

aviflax:
> I strongly suggest moving the inline "TODO"s scattered through the codebase to issues so that others can help contribute to this project.

Good idea, I’ll do so in a follow-up PR. Thanks!

aviflax:
I’ve decided to open a follow-up PR with those commits that will raise test coverage; the gains are fairly small but the changes are non-trivial.

PR #66, opened 2018-09-04
Trifurcate CI Job

aviflax:
For faster builds.

Also a few tweaks, of course — who can resist?


aviflax:
Yeah, that comment is actually in the file 3 times... I’ll fix it. Thank you!

PR #70, opened 2018-09-06
sample function to process a file

AndreaCrotti:
Just in case you think it might be useful @aviflax 
I guess it should probably live in a different namespace though since it's not really related with clipboards..

### DIFF
--- a/tool/src/main/fc4/integrations/structurizr/express/clipboard.clj
+++ b/tool/src/main/fc4/integrations/structurizr/express/clipboard.clj
@@ -4,8 +4,7 @@
              :as ed
              :refer [probably-diagram-yaml? process-file]])
   (:import [java.awt Toolkit]
-           [java.awt.datatransfer DataFlavor StringSelection])
-  (:refer-clojure :exclude [slurp spit]))
+           [java.awt.datatransfer DataFlavor StringSelection]))
 
 ; Suppress the Java icon from popping up and grabbing focus on MacOS.
 ; Found in a comment to this answer: https://stackoverflow.com/a/17544259/7012
@@ -21,7 +20,7 @@
 (def string-flavor (DataFlavor/stringFlavor))
 
 ;; based on code found at https://gist.github.com/Folcon/1167903
-(defn slurp []
+(defn cb-slurp []
   (try
     ;; We’re gonna check twice for the contents being a string because of
     ;; possible race conditions.
@@ -31,7 +30,7 @@
           (.getTransferData transferable string-flavor))))
     (catch java.lang.NullPointerException e nil)))
 
-(defn spit
+(defn cb-spit
   [s]
   {:pre [(string? s)]}
   (.setContents (clipboard) (StringSelection. s) nil))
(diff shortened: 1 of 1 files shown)


aviflax:
Thanks!

I agree, it should go in a different namespace. Probably `fc4c.io`.

Also, I prefer PRs to be as focused as possible. So I’d prefer renaming functions in this namespace to be done in a separate PR from the one that adds such a function.

AndreaCrotti:
It is related @aviflax, because without that change I would have to call `slurp` with `clojure.core/slurp` for example.
So if I move it to a different file then it's not related, but at the moment it was.

sgerrand:
Somewhat of a tangent, but what's the idea behind `fc4c.io` as a namespace? Is there a link to the Indian Ocean that I've previously missed? Would `fc4c.org` make more sense? (💯 ❓s)

aviflax:
@AndreaCrotti wrote:

> It is related @aviflax, because without that change I would have to call `slurp` with `clojure.core/slurp` for example.
So if I move it to a different file then it's not related, but at that moment it was.

True, true. I think I meant to address the scenario wherein the function is moved to a different namespace. In that case, I’d advocate for removing the other changes from the PR so it’s more focused.

----

@sgerrand wrote:

> Somewhat of a tangent, but what's the idea behind `fc4c.io` as a namespace? Is there a link to the Indian Ocean that I've previously missed? Would `fc4c.org` make more sense? (💯 ❓s)

In the Clojure community, it’s idiomatic to name namespaces after a domain, subdomain, domain concept, feature, or area of functionality ([Sierra, 2011](https://stuartsierra.com/2011/08/08/clojure-namespaces); [Hagleberg, 2010](https://groups.google.com/d/msg/clojure/lo6QZiI8Ois/mfmCeSG0K5EJ)). ([Reverse domain notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) is hardly ever used in Clojure projects.) In this case, `io` refers to I/O, Input/Output.

The idea here is to constrain non-pure functions to this one namespace, so as to keep them isolated from the rest of the pure functions. This is my attempt to respect the community’s general consensus that it’s best to keep side effects “at the edges” ([Corfield, 2017](https://coderanch.com/t/685451/paradigms/Functional-Programming-Java-put-Side#3217424)) so most of an app is easier to debug, understand, test, etc. This makes sense to me because pure and non-pure functions lend themselves to notably different approaches of testing and debugging.

sgerrand:
> In this case, io refers to I/O, Input/Output.

Got it, thanks!

aviflax:
@AndreaCrotti if you feel like moving this function to `fc4.io` and adding a test then I’d be inclined to accept this… up to you of course.

AndreaCrotti:
I'm not sure it's worth tbh, I originally did this thinking that with fc4c I could simply generate the diagram directly, in which case I would have wanted a simple way to pass an input and get an output file.
Since it's not the case it probably doesn't help much to have that in..

AndreaCrotti:
And about the focus of PRs @aviflax up to you, but I think in my experience being too strict about these kind of rules only harm the quality of the project overall, since if you expect others to create a separate PR for every small thing they think could be improved they simply won't do it.
(and this is particularly true whenever we have a more strict process like ST which means creating PRs alone is not enough)

I agree they should at least be done in separate commits though.

aviflax:
> I'm not sure it's worth tbh, I originally did this thinking that with fc4c I could simply generate the diagram directly, in which case I would have wanted a simple way to pass an input and get an output file.

Aha, I see, yeah, that makes sense.

Apropos to that, I just posted a draft of [Enhancement Proposal EP02: Automated Rendering](https://github.com/FundingCircle/fc4-framework/pull/74) — if you have a few minutes I’d love your feedback on that, and/or on [its relative priority to other potential enhancements](https://github.com/FundingCircle/fc4-framework/issues/72).

> …I think in my experience being too strict about these kind of rules only harm the quality of the project overall, since if you expect others to create a separate PR for every small thing they think could be improved they simply won't do it.

Good food for thought, I’ll definitely consider this. Thank you!

PR #80, opened 2018-10-29
Node Renderer

aviflax:
This changeset adds a new `renderer` “module” — really just a directory — to fc4-tool. This directory contains `render.js` and everything it needs to function (`package.json` etc). This script implements the approach recommended in #74: it uses Node, Puppeteer, and Chromium to accept a diagram’s YAML representation via stdin, load Structurizr Express (SE) via the Internet, “paste” the diagram YAML representation into SE, trigger the SE “export” feature, and then write the diagram’s PNG image to stdout.

This generally works for me when I test it (manually, on my Mac) using this command from within the `renderer` directory:

```shell
cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js | open -a Preview.app -f
```

Note that before that’ll work you’ll need to run `npm install` from within `renderer`.

This changeset is obviously incomplete as it includes no automated tests and no documentation updates. I’ve got that work completed but I decided to post them as follow-up PRs so that each change is more focused and thus easier to review effectively.

Specifically, this PR will be followed by these PRs:

1. One that adds an automated test and a better CLI for the script, both implemented in Clojure
1. One that gets the automated test working in our CI environment
   1. and also makes it easier to use more or less the same Docker setup locally that we use in CI
1. One that updates the docs 

That’s all I’ve got planned for now, but we’ll see.

PR #74, opened 2018-10-09
Enhancement Proposal EP02: Add Automated Rendering to fc4-tool

aviflax:
Please see #72 for context.

PR #83, opened 2018-10-31
Rendering on Alpine

aviflax:
This is part of the implementation for #74 and a follow-up to #80 and #82.

#82 added a Clojure wrapper and an automated test for the Node renderer added in #80, but wasn’t able to build successfully in CI as the CI build environment needed some updates.

This PR makes those updates, and also leverages those updates to enable running all the tests locally in Docker using the same exact environment and configuration as the CI build environment.

This PR does not include the necessary updates to the docs. Those are coming shortly in a follow-up PR. I kept them separate because _this_ PR is already quite big and I didn’t want this to be too difficult too review.

### Before Merging

* [ ] ~~#82 must merge~~
* [ ] ~~this PR must be re-pointed to master~~

I’ve realized the above steps won’t work. For an explanation please see the note at the end of the description of #82.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,8 @@ jobs:
        - run:
           name: Download Node Dependencies (if necessary)
           working_directory: ~/project/tool/renderer
+          environment:
+            PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
           command: npm install
 
        - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,15 +39,11 @@ jobs:
 
        - run:
           name: Download Clojure Dependencies (if necessary)
-          # Clojure automatically downloads deps if necessary
-          command: clojure -R:test:test/coverage:lint -Stree
+          command: bin/download-clojure-deps
 
        - run:
           name: Download Node Dependencies (if necessary)
-          working_directory: ~/project/tool/renderer
-          environment:
-            PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
-          command: npm install
+          command: bin/download-node-deps
 
        - save_cache:
            name: Save Clojure Dependencies (if changed)
@@ -61,7 +57,7 @@ jobs:
            name: Save Node Dependencies (if changed)
            key: node-deps-{{checksum "renderer/package.json"}}
            paths:
-           - ~/project/tool/renderer/node_modules
+           - renderer/node_modules
 
    tool_test:
      <<: *tool_job_defaults
@@ -77,10 +73,7 @@ jobs:
 
        - run:
           name: Run tests and measure test coverage
-          # The max heap size is set to 2GB because I’ve seen OOM errors at 1GB and below. (JDK 8
-          # defaults to setting the max heap to ¼ of the total RAM, and CI containers frequently have
-          # <= 4GB RAM.)
-          command: clojure -J-Xmx2g -A:test:test/coverage
+          command: bin/tests-with-coverage
 
        - store_test_results:
           path: target/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,13 +10,19 @@ version: 2
 tool_job_defaults: &tool_job_defaults
   working_directory: ~/project/tool
   docker:
-    - image: clojure:tools-deps-alpine
+    - image: quay.io/aviflax/clojure-node:clojure-1.9.0.394_node-10.12
 
-restore_cache_keys: &restore_cache_keys
+restore_clj_cache_keys: &restore_clj_cache_keys
   keys:
-  - deps-{{ checksum "deps.edn" }}
+  - clj-deps-{{ checksum "deps.edn" }}
   # fallback to using the latest cache if no exact match is found
-  - deps-
+  - clj-deps-
+
+restore_node_cache_keys: &restore_node_cache_keys
+  keys:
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ version: 2
 jobs:
    tool_deps:
      docker:
-       - image: clojure:tools-deps-alpine
+       - image: quay.io/aviflax/clojure-node-chromium:clojure-1.9.0.394_node-10.12
```
aviflax: Is it OK to have this project rely on an image in my personal Quay account, as opposed to the Funding Circle account? I was thinking maybe this is OK, because the Dockerfile for the image is right here in this repo (added in this branch) in `.circleci/images/tool/Dockerfile` so if the image ever needs to be updated and I’m not available to do so, the new version of the image could be pushed to Funding Circle’s account, and then this line could be changed pretty easily.

sgerrand: Let's change it to `quay.io/fundingcircle/clojure-node-chromium`.

aviflax: Sure, will do, thanks!

sgerrand: I've created the Docker repository for you and it's set up to build from `.circleci/images/tool/Dockerfile` automatically.

aviflax: Thank you!

aviflax: Fixed in f0dcb92

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ version: 2
 jobs:
    tool_deps:
      docker:
-       - image: clojure:tools-deps-alpine
+       - image: quay.io/fundingcircle/clojure-node-chromium@sha256:84a563c327d900cbfacbeabf41f941a2395ca55835135f639e8b5d656b8d7c80
```
sgerrand: 💭 Suggest using a tag instead of the image hash.

.circleci/images/tool/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+LABEL maintainer="Avi Flax <avi.flax@fundingcircle.com>"
+LABEL description="Node, Clojure, and Chromium on Alpine"
+
+ENV JRE_VERSION=8.181.13-r0
+ENV CLOJURE_VERSION=1.9.0.394
+
+WORKDIR /tmp
+
+RUN apk update
```
sgerrand: 🙋 Running this command negates the `--no-cache` flag used below.

https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Update_the_Package_list

sgerrand: Using `upgrade` means that cache files are added to the local filesystem, which is probably not what you want to do. If you're trying to _upgrade_ the existing packages, then you can pass the `--upgrade` flag to the `aok add` command below.

.circleci/images/tool/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+# Dependencies of Clojure.
+# NB: the OpenJDK image published by Docker the company does a few things beyond
+# just apk adding the openjdk-jre package; those things might also be necessary;
+# not sure. See
+# https://github.com/docker-library/openjdk/blob/master/8/jre/alpine/Dockerfile
+RUN apk -q --no-progress add --no-cache openjdk8-jre=$JRE_VERSION bash curl
```
sgerrand: 🙋 Suggest placing flags in front _or_ after the invoked command, not both.

tool/Dockerfile.test
```diff
@@ -0,0 +1,27 @@
...
+COPY deps.edn ./
+COPY bin/download-clojure-deps bin/
+RUN bin/download-clojure-deps
+
+COPY renderer/package*.json ./renderer/
+COPY bin/download-node-deps bin/
+RUN bin/download-node-deps
+
+# rlwrap is handy in case we need to run a REPL in a container
+RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
```
sgerrand: 🙋 As an alternative approach, you could specify the repository used during `apk add` by using the `--repository` flag.

e.g.

`apk add --no-cache --repository=http://nl.alpinelinux.org/alpine/edge/testing rlwrap`

.circleci/config.yml
aviflax: I haven’t seen such issues with Alpine Linux specifically. But I’ve seen such issues in general — an unintended/unplanned dependency upgrade — causing lots of problems. So I prefer to be cautious and pin as much as possible, as specifically as possible.

Even though we control the image tags (and this might just be the lack of depth in my Docker expertise) I’d be concerned if we have occasion to rebuild our base image and then at that occasion, the `node` base image has _changed_ — then the whole situation is non-deterministic. I don’t like that, so I’m going to use the hash. I’ll just move it to where it should be. Thanks!

aviflax: Addressed in 3370afc — thanks!

.circleci/images/tool/Dockerfile
aviflax: Addressed in 3370afc — thanks!

tool/Dockerfile.test
aviflax: Addressed in 3370afc — thanks!

.circleci/images/tool/Dockerfile
aviflax: Addressed in 3370afc — thanks!

PR #94, opened 2018-11-14
Support recent/current versions of Chrome/Chromium

### LINKED ISSUE #92: Renderer only works with Chromium/Chrome 68
I discovered today that the version of Puppeteer we’re currently using (1.4.0, because that’s the latest version supported on Alpine, because Alpine [only has Chrome 68](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#tips) and we’re using Alpine for our tests) no longer works with the latest Chromium/Chrome (72) — it requires Chromium/Chrome 68 (it _might_ work with 69–71 but they’re explicitly not supported).

This is annoying, and we’re going to have to do something about it soon.

See also #91.

aviflax:
This switches our test+build containers from Alpine to Debian Stretch,
which has a more recent version of Chromium. This allows us to upgrade
to a less ancient version of Puppeteer; this less ancient Puppeteer is
compatible with modern Chrome/Chromium. Therefore users no longer need
to have the ancient+obsolete Chrome/Chromium 68 installed on their
systems, which was unlikely, annoying, and insecure.

This also upgrades from Node 10 to Node 11, because why not 😬

Fixes #92

### DIFF
--- a/.circleci/images/tool/Dockerfile
+++ b/.circleci/images/tool/Dockerfile
@@ -1,32 +1,21 @@
-# I wish the “official” Node images included a tag to fix the version of Alpine
-# used, but they don’t. Therefore we’re specifing a specific image hash to use
-# rather than a tag.
-# This base image at this hash uses Node 10.12 and Alpine 3.8:
-FROM node@sha256:1e3e3e7ffc965511c5d4f4e90ec5d9cabee95b5b1fbcd49eb6a2289f425cf183
+FROM node:11.1.0-stretch
 
 LABEL maintainer="Avi Flax <avi.flax@fundingcircle.com>"
-LABEL description="Node, Clojure, and Chromium on Alpine"
+LABEL description="Node 11, Clojure 1.9, and Chromium 70 on Debian Stretch"
 
-ENV JRE_VERSION=8.181.13-r0
+ENV JRE_VERSION=8u181-b13-2~deb9u1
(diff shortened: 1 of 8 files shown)


.circleci/images/tool/Dockerfile
```diff
@@ -1,32 +1,21 @@
-# I wish the “official” Node images included a tag to fix the version of Alpine
-# used, but they don’t. Therefore we’re specifing a specific image hash to use
-# rather than a tag.
-# This base image at this hash uses Node 10.12 and Alpine 3.8:
-FROM node@sha256:1e3e3e7ffc965511c5d4f4e90ec5d9cabee95b5b1fbcd49eb6a2289f425cf183
+FROM node:11.1.0-stretch
```
sgerrand: 🙋 I'd stay on the LTS release.

aviflax:
> Because Node v10 is a LTS release? That's a good enough argument to me for staying on it.

Absolutely! Thanks!

.circleci/images/tool/Dockerfile
aviflax: Agreed, fixing now. Thank you!

sgerrand:
LGTM otherwise! 😁

.circleci/images/tool/Dockerfile
aviflax: Done in 03ad139 — thanks!

PR #95, opened 2018-11-15
Fix packaged renderer + two other small test/build/package improvements

aviflax:
I recommend reviewing each commit individually; they’re fully decoupled. Thanks!

### Before Merging

* [x] #98 must merge
* [x] re-point this to master

### DIFF
--- a/.circleci/images/tool/Dockerfile
+++ b/.circleci/images/tool/Dockerfile
@@ -3,19 +3,25 @@ FROM node:10.13.0-stretch
 LABEL maintainer="Avi Flax <avi.flax@fundingcircle.com>"
 LABEL description="Node 10 (LTS), Clojure 1.9, and Chromium 70 on Debian Stretch"
 
-ENV JRE_VERSION=8u181-b13-2~deb9u1
-ENV CLOJURE_VERSION=1.9.0.394
-ENV CHROMIUM_VERSION=70.0.3538.67-1~deb9u1
-
 WORKDIR /tmp
 
 # Install dependencies of the Renderer and Clojure
+#
+# Version numbers are hard-coded because if we use environment variables, the
+# ENV statements prevent all subsequent layers from being cached, because
+# they’re considered deterministic. I’d prefer that the version numbers were
+# stated more declaratively, all in one place, and there’s probably a way to do
+# that without ENV but I couldn’t find a straightforward way to do so.
+#
 # We’re including rlwrap just because it’s handy in case we need to run a REPL
 # in a container, for debugging.
 RUN apt-get update -yq
-RUN apt-get install -yq chromium=$CHROMIUM_VERSION openjdk-8-jre=$JRE_VERSION rlwrap
+RUN apt-get install -yq \
+      chromium=70.0.3538.67-1~deb9u1 \
+      openjdk-8-jre=8u181-b13-2~deb9u1 \
+      rlwrap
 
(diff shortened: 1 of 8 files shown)


.circleci/images/tool/Dockerfile
```diff
@@ -3,19 +3,25 @@ FROM node:10.13.0-stretch
...
+# ENV statements prevent all subsequent layers from being cached, because
+# they’re considered deterministic. I’d prefer that the version numbers were
+# stated more declaratively, all in one place, and there’s probably a way to do
+# that without ENV but I couldn’t find a straightforward way to do so.
+#
 # We’re including rlwrap just because it’s handy in case we need to run a REPL
 # in a container, for debugging.
 RUN apt-get update -yq
-RUN apt-get install -yq chromium=$CHROMIUM_VERSION openjdk-8-jre=$JRE_VERSION rlwrap
+RUN apt-get install -yq \
```
sgerrand: 💭 What about:
```docker
RUN JRE_VERSION=8u181-b13-2~deb9u1 \
    CLOJURE_VERSION=1.9.0.394 \
    CHROMIUM_VERSION=70.0.3538.67-1~deb9u1 \
    apt-get install -yq \
      chromium=$CHROMIUM_VERSION openjdk-8-jre=$JRE_VERSION rlwrap
```

sgerrand: and
```
RUN CLOJURE_VERSION=1.9.0.394 \
    wget -q https://download.clojure.org/install/linux-install-$CLOJURE_VERSION.sh \
      && chmod +x linux-install-$CLOJURE_VERSION.sh \
      && ./linux-install-$CLOJURE_VERSION.sh
```

aviflax: Tried that, ran into trouble because Docker thinks those variable references are to Docker variables, and replaces them with nothing. I also tried escaping the $ but had trouble with that too. Even tried passing it all as a string to bash -c but no dice. At a certain point I decided to cut my losses. I could try again to figure something out if you think it’s important.

tool/renderer/package.json
```diff
@@ -1,10 +1,14 @@
 {
-  "name": "fc4-se-renderer",
+  "name": "render",
```
sgerrand: 🙋 A npm package called [`render` already exists](https://www.npmjs.com/package/render). If the package name change is desired then I suggest setting `"private": true` in this file to prevent this package being accidentally published to the npm registry.

aviflax: Good tip, will do, thanks!

.circleci/images/tool/Dockerfile
sgerrand: I don't think it's that important. :bow:

Probably worth moving the Docker image build out of Quay and into this CircleCI job at some point.

aviflax: > I don't think it's that important. 🙇

😎
 
> Probably worth moving the Docker image build out of Quay and into this CircleCI job at some point.

I’ve opened up #96 to discuss+track this.

tool/renderer/package.json
aviflax: Done in b68749c — thanks!

aviflax:
This is blocked because the tests cannot pass because Structurizr Express has been changed so automated rendering is broken. I should have a fix up today in a new PR that’ll have to merge before this one.

PR #98, opened 2018-11-20
Fix + improve automated rendering

aviflax:
Simon Brown, the owner and maintainer of Structurizr Express, has made some changes to its code that broke our renderer.

Thankfully, he also published [a new example script](https://github.com/structurizr/puppeteer) that demonstrates the optimal way to render and export diagrams via Structurizr Express; this enabled me to both fix our renderer but also improve it.

So this adapts the renderer to the changes but also makes our code simpler and faster.

This required no test changes; the (Clojure) test that invokes the renderer is passing.

### DIFF
--- a/tool/renderer/render.js
+++ b/tool/renderer/render.js
@@ -50,86 +50,40 @@ function puppeteerOpts(debugMode) {
   };
 }
 
-function abit(ms) {
-  log.next(`pausing ${ms}ms`);
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
-
 async function loadStructurizrExpress(browser, url) {
   log.next(`loading Structurizr Express from ${url}`);
-  const mainPage = await browser.newPage();
-  await mainPage.goto(url, {'waitUntil' : 'networkidle2'});
-  return mainPage;
+  const page = await browser.newPage();
+  await page.goto(url);
+
+  // I copied this step from
+  // https://github.com/structurizr/puppeteer/blob/d8b625aef77b404de42199a1ff9a9f6730795913/export-express-diagram.js#L24
+  await page.waitForXPath("//*[name()='svg']");
+
+  return page;
 }
 
 async function setYamlAndUpdateDiagram(page, diagramYaml) {
   log.next('setting YAML and updating diagram');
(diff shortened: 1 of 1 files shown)


tool/renderer/render.js
```diff
@@ -142,20 +96,29 @@ async function closeBrowser(browser, debugMode) {
...
   log.next('launching browser');
   const opts = puppeteerOpts(debugMode);
   const browser = await puppeteer.launch(opts);
 
-  const imageBuffer = await render(theYaml, browser, url, debugMode);
+  const imageBuffer = await render(preppedYaml, browser, url, debugMode);
   closeBrowser(browser, debugMode);
 
   process.stdout.write(imageBuffer);
 }
```
fussybeaver: should there be some error handling somewhere? There's a lot of file operations, opening browsers, presumably in a different process. Not sure what is likely to fail here - is there any other sanitization done on the arguments to the main method ?

tool/renderer/render.js
```diff
@@ -50,86 +50,40 @@ function puppeteerOpts(debugMode) {
...
-  const newPagePromise = new Promise(resolve => {
-    mainPage.browser().once('targetcreated', target => resolve(target.page()));
-  });
-
-  await abit(200);
-  await mainPage.evaluate(() => {
-    Structurizr.diagram.exportCurrentView(1, true, false, false, false);
+async function exportDiagram(page) {
+  const diagramImageBase64DataURI = await page.evaluate(() => {
+    return structurizr.scripting.exportCurrentDiagramToPNG();
```
fussybeaver: This doesn't actually do any manipulation of the image data in javascript ?

aviflax: Correct. The purpose of the renderer — well honestly it’s not _really_ a renderer! Structurizr Express is doing the rendering; this script is more of an adapter that uses Structurizr Express to render YAML files to PNG files. Maybe I should rename the script?

tool/renderer/render.js
aviflax: Yes, there should definitely be error handling. I do need to add it ASAP — I’ve created #99 to track that. Thank you!

PR #106, opened 2018-11-26
Fix rendering and improve the renderer’s error-handling

### LINKED ISSUE #103: Automated rendering seems to be broken
As reported by @stuart-robinson:

```
cat global/originations_platform/orginations_platform_02_container.yaml | fc4-render --verbose
launching browser...
loading Structurizr Express from https://structurizr.com/express?autoLayout=false...
setting YAML and updating diagram...
calling export function...
(node:43666) UnhandledPromiseRejectionWarning: Error: Evaluation failed: TypeError: Cannot read property 'crop' of undefined
    at Structurizr.DiagramScripting.exportCurrentDiagramToPNG (https://structurizr.com/static/js/structurizr-scripting-diagram.js?v=1380:13:59)
    at __puppeteer_evaluation_script__:2:34
    at ExecutionContext.evaluateHandle (/snapshot/root/project/tool/renderer/node_modules/puppeteer-core/lib/ExecutionContext.js:106:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:43666) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:43666) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I suspect Structurizr Express may have changed. I will investigate shortly.

### LINKED ISSUE #104: When rendering fails the renderer just hangs
This is bad — it should fail fast with a clear error message!

As mentioned in #103.

<details>
<summary>Example 1</summary>
<pre><code>
cat global/originations_platform/orginations_platform_02_container.yaml | fc4-render --verbose
launching browser...
loading Structurizr Express from https://structurizr.com/express?autoLayout=false...
setting YAML and updating diagram...
calling export function...
(node:43666) UnhandledPromiseRejectionWarning: Error: Evaluation failed: TypeError: Cannot read property 'crop' of undefined
    at Structurizr.DiagramScripting.exportCurrentDiagramToPNG (https://structurizr.com/static/js/structurizr-scripting-diagram.js?v=1380:13:59)
    at __puppeteer_evaluation_script__:2:34
    at ExecutionContext.evaluateHandle (/snapshot/root/project/tool/renderer/node_modules/puppeteer-core/lib/ExecutionContext.js:106:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:43666) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:43666) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
</code></pre>
</details>

<details>
<summary>Example 2</summary>
<pre><code>
$ cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js --verbose | open -a Preview.app -f
launching browser...
loading Structurizr Express from https://structurizr.com/express?autoLayout=false...
(node:25642) UnhandledPromiseRejectionWarning: TimeoutError: Navigation Timeout Exceeded: 30000ms exceeded
    at Promise.then (/Users/avi.flax/dev/fc4-framework/tool/renderer/node_modules/puppeteer-core/lib/NavigatorWatcher.js:94:21)
(node:25642) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:25642) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
</code></pre>
</details>

aviflax:
Please see the individual commits.

Fixes #103
Fixes #104

### DIFF
--- a/tool/renderer/render.js
+++ b/tool/renderer/render.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 // NB: here’s how I’ve been testing this on my Mac:
-// renderer $ cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js --verbose | open -a Preview.app -f
+// renderer $ cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js | open -a Preview.app -f
 
 const dataUriToBuffer = require('data-uri-to-buffer');
 const {existsSync, readFileSync} = require('fs');
@@ -86,6 +86,19 @@ async function render(diagramYaml, browser, url, debugMode) {
   return imageBuffer;
 }
 
+// On success: returns a Puppeteer browser object
+// On failure: logs an error then returns null
+async function launchBrowser(debugMode) {
+  try {
+    const opts = puppeteerOpts(debugMode);
+    log.next('launching browser');
+    return await puppeteer.launch(opts);
+  } catch (err) {
+    console.error(`Could not launch browser: ${err}\n${err.stack}`);
+    return null;
+  }
+}
+
 async function closeBrowser(browser, debugMode) {
   if (debugMode) {
     log.next('DEBUG MODE: leaving browser open; script may be blocked until the browser quits.');
@@ -108,15 +121,25 @@ async function main(url, debugMode) {
   const rawYaml = readFileSync("/dev/stdin", "utf-8");
   const preppedYaml = prepYaml(rawYaml);
 
-  log.next('launching browser');
-  const opts = puppeteerOpts(debugMode);
-  const browser = await puppeteer.launch(opts);
+  // This is outside of the try block so that the binding will be visible to
+  // both the try block below and the finally block, because if an error occurs
+  // it’s really important to close the browser; if we don’t then the program
+  // will hang and not exit, even though rendering failed.
+  const browser = await launchBrowser(debugMode);
+
+  if (!browser) {
+    // An error message will have been printed out by launchBrowser
+    process.exitCode = 1;
+    return;
+  }
 
   try {
     const imageBuffer = await render(preppedYaml, browser, url, debugMode);
     process.stdout.write(imageBuffer);
   } catch (err) {
-    console.error(err);
+    console.error(
+      `RENDERING FAILED\n${err}\n${err.stack}\nPrepped YAML:\n${preppedYaml}`
+    );
     process.exitCode = 1;
   } finally {
     closeBrowser(browser, debugMode);
--- a/tool/renderer/render.js
+++ b/tool/renderer/render.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 // NB: here’s how I’ve been testing this on my Mac:
-// renderer $ cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js | open -a Preview.app -f
+// renderer $ cat ../test/data/structurizr/express/diagram_valid_cleaned.yaml | time ./render.js --verbose | open -a Preview.app -f
 
 const dataUriToBuffer = require('data-uri-to-buffer');
 const {existsSync, readFileSync} = require('fs');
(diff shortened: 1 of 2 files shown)


tool/renderer/render.js
```diff
@@ -111,12 +112,17 @@ async function main(url, debugMode) {
   const opts = puppeteerOpts(debugMode);
   const browser = await puppeteer.launch(opts);
```
fussybeaver: Could this fail if the wrong browser version is used (or if Chromium is missing)?

tool/renderer/render.js
```diff
@@ -111,12 +112,17 @@ async function main(url, debugMode) {
...
 
-  const imageBuffer = await render(preppedYaml, browser, url, debugMode);
-  closeBrowser(browser, debugMode);
-
-  process.stdout.write(imageBuffer);
+  try {
+    const imageBuffer = await render(preppedYaml, browser, url, debugMode);
+    process.stdout.write(imageBuffer);
+  } catch (err) {
+    console.error(err);
```
fussybeaver: Might be worth outputting information like which `url` failed, and what the `preppedYaml` looks like (if appropriate).

tool/renderer/render.js
aviflax: Hypothetically, sure. But in practice I don’t think I’ve seen that behavior from Puppeteer. My guess is that rather than checking the version of Chromium/Chrome being used and failing or warning if the version is unsupported, it takes a best-effort approach and tries to proceed, hoping for the best — any given library call may succeed or fail, I suppose, and I’ve seen individual calls fail when an incompatible version of Chromium/Chrome is used.

All that said, certainly launching the browser could fail due to incompatible versions or any other number of reasons, and yes, this line should probably also be included in a try/catch block.

If you don’t mind, I’d rather not delay this PR in order to make that additional improvement, because currently rendering is straight-out broken. Perhaps instead I could create a new issue to improve the error-handling more systematically?

tool/renderer/render.js
aviflax: Good suggestions, thanks!

WRT the former, I don’t think that’s necessary because the URL will be effectively a constant; it’s always Structurizr Express. The only reason it’s a variable/argument is because at one point I was experimenting with embedding Structurizr Express into the renderer and loading it from the filesystem — but that turned out to be a non-viable option. Your suggestion makes me realize that the value should in fact be a module-level constant, in other words it shouldn’t be an argument, because it being an argument implies that it will vary. I’ll make that change, but I’d rather do it in a follow-up PR if that’s OK.

WRT the latter, yes, I totally agree. I’ll try to get that in shortly. Thanks!

tool/renderer/render.js
aviflax: Ah never mind, I’m just doing this now.

tool/renderer/render.js
aviflax: Ah never mind, I’m just doing this now.

aviflax: Please see 1ea17f1 and 50a3d54.

tool/renderer/render.js
aviflax: Please see 1ea17f1. Thanks!

tool/renderer/render.js
```diff
@@ -78,13 +81,26 @@ async function exportDiagram(page) {
...
 
+// On success: returns a Puppeteer browser object
+// On failure: logs an error then returns null
+async function launchBrowser(debugMode) {
+  try {
+    const opts = puppeteerOpts(debugMode);
+    log.next('launching browser');
+    return await puppeteer.launch(opts);
+  } catch (err) {
+    console.error(`Could not launch browser: ${err}\n${err.stack}`);
```
fussybeaver: Could emit the options passed to puppeteer.

aviflax: Agreed, but don’t have time right now to make more improvements.

PR #107, opened 2018-11-26
Refactor CircleCI Configuration

aviflax:
The main change here is to separately download+cache deps for testing+packaging (d5fc8a1). I recommend reviewing each commit individually. Thanks!


.circleci/config.yml
```diff
@@ -10,114 +10,46 @@ version: 2
...
+          key: test-deps-{{checksum "deps.edn"}}-{{checksum "renderer/package.json"}}-{{checksum "bin/download-test-deps"}}
+          paths: [.cpcache, ~/.m2, ~/.gitlibs, renderer/node_modules]
+       - store_test_results: {path: target/test-results}
+       - store_artifacts: {path: target/coverage}
        - run:
           name: Upload test coverage report to Codecov
           command: bash <(curl -s https://codecov.io/bash)
 
-   tool_lint:
+   tool_clj_lint:
```
fussybeaver: Is it worth adding a javascript linter ?

aviflax: Yes, that’s my plan — that’s why I changed the name of this job to make it more specific. I’ll do this in a follow-up PR.

tool/bin/docker-pkg-run
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
+# Run commands using ../Dockerfile.pkg
+# e.g. from <repo-root>/tool:  bin/docker-pkg-run bin/pkg-all
+
+set -ex
```
fussybeaver: Is `-x` needed in all of these scripts ? It seems easy enough to follow when I look at the CircleCI job...

aviflax: Well, the scripts are useful outside of CI as well. I like the convenience of scripts, and how they are sort-of like executable documentation, and provide leverage, but I dislike when they’re mysterious. For scripts like these that are likely to be used only occasionally by developers, I like to set `-x` so as to make it explicit and super clear what’s happening and how.

.circleci/config.yml
aviflax: Opened #112 — thanks!

PR #111, opened 2018-11-27
Enhancement Proposal EP05: Unified Authoring Workflow

aviflax:
I look forward to your comments!

### DIFF
--- a/proposals/ep05-new-editing-workflow/ep05-new-editing-workflow.md
+++ b/proposals/ep05-new-editing-workflow/ep05-new-editing-workflow.md
@@ -15,7 +15,7 @@
   </tr>
   <tr>
     <th>Status:</th>
-    <td>Draft</td>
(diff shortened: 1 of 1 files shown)


aviflax:
I demoed the new workflow for @99-not-out and got positive feedback, he thinks it’s solid and compelling. (He even downloaded [the prerelease](https://github.com/FundingCircle/fc4-framework/releases/tag/prerelease_ep05-spike) to try it out himself.)

I also showed Matt this proposal and suggested that the single-file workflow is way too complex; hard to explain; the clipboard watching/pasting/copying feels hard to predict, sort of like “spooky action at a distance”; and that there’s too much cognitive load involved in even having two different workflows. I suggested that I remove the current single-file workflow with the clipboard integration and make the multi-file workflow the only workflow. Matt agreed, so I’m going to do that.

Thanks Matt!

Issue #97, opened 2018-11-20
usage of the release


bertvannuffelen:
Hi,

I wanted to test the release. I downloaded and unpacked it on my linux box.
and then copied one of the tests files and tried to render it. But the tool gets stuck and does not produce any result.

``` 
cp ../../tool/test/data/model\ \(valid\)/systems/division-a/middle.yaml .
./fc4-render middle.yaml 
```

What do I miss? kr Bert

aviflax:
Hi Bert, my apologies! I have a few fixes for this that should be merged soon (and each merge results in a new release with new download packages). I’ll let you know once they’re merged and the new release is ready. In the meantime, you’re welcome to try [this prerelease](https://github.com/FundingCircle/fc4-framework/releases/tag/prerelease_publish-prereleases) — the Mac download, at least, works on my machine ;)

If you’re curious about the problems their solutions, feel free to check out #95 and #98.

Also, a few notes on your test commands:

* The YAML file you’re copying, `middle.yaml`, is not an FC4 diagram
  * (which, at the moment, is synonomous with a [Structurizr Express](https://structurizr.com/express) diagram)
  * it’s a file describing a single software system; part of a set of files that together define a model of an entire system landscape, or maybe even multiple landscapes. This is a new DSL that I’m working on, but isn’t yet fully supported. (I’d love your feedback on #73 if you’re interested.)
  * As an alternative, you can use `<project_root>/tool/test/data/structurizr/express/diagram_valid_messy.yaml`
* `fc4-render` isn’t meant to be used directly; it’s meant to be invoked via the main `fc4` command, which you’d invoke like: `fc4 render path/to/diagram.yaml`
  * If you’d still like to run `fc4-render` directly, for debugging or maybe for the shorter execution times, you’ll need to pipe input to it and save its output yourself, like: `cat path/to/diagram.yaml | fc4-render > diagram.png`

aviflax:
Hi Bert, those fixes have been merged and there’s [a fresh new release](https://github.com/FundingCircle/fc4-framework/releases/tag/master_2018-11-21_650) — I’d love to hear whether it works for you. If it doesn’t, please share the command you’re using, your OS, your Java install, and anything else you think might help me reproduce.

bertvannuffelen:
Hi Avi,

thanks for the update. There seem to be still some issues left:

```
root@2284d6376a66:/data/fc4# fc4 render diagram_valid_cleaned.yaml 
bash: fc4: command not found
[...]

PR #114, opened 2018-11-27
Add example Dockerfile that shows how to install fc4-tool and render a diagram

aviflax:
As discussed in #97

### DIFF
--- a/tool/doc/examples/docker-install-and-render/Dockerfile
+++ b/tool/doc/examples/docker-install-and-render/Dockerfile
@@ -9,9 +9,9 @@ RUN apt-get -yq update \
 
 # Install fc4-tool
 RUN curl -L -s \
-      https://github.com/FundingCircle/fc4-framework/releases/download/master_2018-11-27_678/fc4-linux.tar.gz \
+      https://github.com/FundingCircle/fc4-framework/releases/download/release_2018-12-11_772/fc4-tool-linux-d021ef9.tar.gz \
       | tar xzv
 
 COPY *.yaml ./
 
-RUN export PATH=$PWD/fc4:$PATH && fc4 render *.yaml
+RUN fc4/fc4 render *.yaml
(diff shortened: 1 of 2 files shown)


tool/doc/examples/docker-install-and-render/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+      && apt-get -yq install --no-install-recommends chromium-browser
+
+# Install fc4-tool
+RUN curl -L -s \
+      https://github.com/FundingCircle/fc4-framework/releases/download/master_2018-11-27_678/fc4-linux.tar.gz \
+      | tar xzv
+
+COPY *.yaml ./
+
+RUN export PATH=$PWD/fc4:$PATH && fc4 render *.yaml
```
sgerrand: 💭 Perhaps set `$PATH` as an `ENV` directive? Not sure if this image is intended to be used solely for generating diagram images via `docker build` or not. If the former, then ignore me. 😸

aviflax: Yeah, I don’t know. I also played with using CMD or ENTRYPOINT. But in the end I decided to stick with this, because this file is really just an example. It’s not really meant to be used as-is, as a tool or whatever; it’s meant to demonstrate how to create a Docker container in which one can use fc4-tool to render diagrams, and I think this does so — one can run a single command, `docker build` and see that it works.

That said, I’d love to hear what @bertvannuffelen thinks, because the impetus to create this example file came from the troubles he reported in #97.

sgerrand: Makes sense, thanks.

aviflax: This was rendered obsolete by 761df0e which updated the FC4 release that’s used; the new release no longer requires that `fc4` and `fc4-render` be in the user’s $PATH.

aviflax:
I haven’t heard back from @bertvannnuffelen but I think this is a useful addition, as does Sasha, so I’m just going to merge this now; we can always revise it later.

PR #115, opened 2018-11-27
[priority: low] Publish distribution packages for all branch builds

aviflax:
as prereleases.

This should be helpful for:

* Debugging — sometimes I need to use a built dist package to reproduce some problem so I can debug it
* Enabling users to try out features that are not 100% ready, so we can get feedback sooner rather than later
* Finding out about breakage in the publishing process sooner rather than later
  * If a commit/build breaks the publishing process somehow, this way we’ll find out

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,11 +95,19 @@ jobs:
            name: Create GitHub Release and upload distribution packages to it
            # Docs for ghr’s options: https://github.com/tcnksm/ghr/#options
            command: |
-             TAG="${CIRCLE_BRANCH}_$(date "+%Y-%m-%d")_${CIRCLE_BUILD_NUM}"
+             if [[ $CIRCLE_BRANCH == "master" ]]; then
+               # Tag has build num suffix in case we do multiple releases on a given day.
+               TAG="release_$(date "+%Y-%m-%d")_${CIRCLE_BUILD_NUM}"
+             else
+               # We need this to be the same across all builds for a branch so we only ever keep the
+               # latest tag+release for a given branch, or we’d have way too many tags+branches.
+               TAG="prerelease_${CIRCLE_BRANCH}"
+             fi
              ~/bin/ghr -t $GITHUB_TOKEN \
                        -u $CIRCLE_PROJECT_USERNAME \
                        -r $CIRCLE_PROJECT_REPONAME \
                        -c $CIRCLE_SHA1 \
+                       $(if [[ ! $CIRCLE_BRANCH == "master" ]]; then echo -prerelease -recreate; fi) \
                        $TAG \
                        /workspace/packages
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -79,23 +79,37 @@ jobs:
        - attach_workspace: {at: /workspace}
        - run: |
            [ "$GITHUB_TOKEN" ] || { echo 'GITHUB_TOKEN is not set!' && exit 1; }
+       - restore_cache: {keys: [ghr_v0.12.0_linux_amd64]}
        - run:
-           ## TODO: cache ghr
            name: Download and unpack ghr
            command: |
+             if [ -f ~/bin/ghr ]; then exit 0; fi
```
sgerrand: 🤔 Checking that this file exists might not be enough to short circuit here, e.g. the file exists but isn't executable. Maybe testing the result of `ghr --version` would be more helpful?

.circleci/config.yml
```diff
@@ -79,23 +79,37 @@ jobs:
...
            command: |
-             TAG="${CIRCLE_BRANCH}_$(date "+%Y-%m-%d")_${CIRCLE_BUILD_NUM}"
-             ./ghr -t $GITHUB_TOKEN \
-                   -u $CIRCLE_PROJECT_USERNAME \
-                   -r $CIRCLE_PROJECT_REPONAME \
-                   -c $CIRCLE_SHA1 \
-                   $TAG \
-                   /workspace/packages
+             if [[ $CIRCLE_BRANCH == "master" ]]; then
+               # Tag has build num suffix in case we do multiple releases on a given day.
```
sgerrand: :+1:

.circleci/config.yml
```diff
@@ -79,23 +79,37 @@ jobs:
...
+               TAG="release_$(date "+%Y-%m-%d")_${CIRCLE_BUILD_NUM}"
+             else
+               # We need this to be the same across all builds for a branch so we only ever keep the
+               # latest tag+release for a given branch, or we’d have way too many tags+branches.
```
sgerrand: :+1:

.circleci/config.yml
aviflax: 🤔 well… I don’t know, this is about caching files, not about testing a tool. If that file exists, then it’s been restored from the cache. And I happen to know that if that file has been restored from the cache, it’s almost certain to be executable. I mean… I guess that seems like quite an unlikely edge case, and maybe a little paranoid of a thing to add to the job definition.

sgerrand: If testing the file is executable is enough, then I'd suggest using `-x` instead of `-f` as the `test` flag.

aviflax: Love it, will do, thanks!

aviflax: Done in 5e7040b — thanks!

PR #128, opened 2018-12-13
Refactor render namespaces

aviflax:
Main goal here was to move most functions from `fc4.cli.render` to `fc4.io.render`.

This is preparatory refactoring for EP05 (#111) for which we’ll soon be adding a new `edit` CLI command that will be implemented in `fc4.cli.edit` and will also need to render YAML files. In other words, the functionality that until now was only in fc4.cli.render will need to be used by two different namespaces under fc4.cli, so we needed that functionality to be in a third namespace. (I’m writing in the future tense because that’s what makes sense in the context of this PR but I’ve actually already implemented `fc4.cli.edit` in a “child” branch.)

I chose to move these functions out of the `cli` source tree and into the `main` source tree because by convention we don’t have tests for the code under `cli` but we do have tests for that under `main`. And since this code will now be shared by more than one ns, and anyway since it was fairly extensive, it should be tested.

Thank you!

### DIFF
--- a/tool/src/main/fc4/io/render.clj
+++ b/tool/src/main/fc4/io/render.clj
@@ -1,0 +1,76 @@
+(ns fc4.io.render
+  "Functions for rendering Structurizr Express diagrams into PNG image files.
+  NB: because these functions are specifically intended for implementing CLI
+  commands, some of them write to stdout/stderr and may call fc4.cli.util/fail
+  (which calls System/exit unless fc4.cli.util/*exit-on-fail* is rebound)."
+  (:require [cognitect.anomalies :as anom]
+            [clojure.java.io :as io :refer [file output-stream]]
+            [clojure.string :as str :refer [split]]
+            [fc4.cli.util :as cu :refer [debug]]
+            [fc4.io :refer [binary-spit]]
+            [fc4.integrations.structurizr.express.render :as r])
+  (:import [java.io FileNotFoundException]))
+
+(defn fail
+  [file-path msg]
+  (cu/fail (str "Error rendering " file-path ": " msg)))
+
+(defn read-text-file
+  [path]
+  (try (slurp path)
+       (catch FileNotFoundException _ (fail path "file not found"))
+       (catch Exception e (fail path (.getMessage e)))))
+
+(defn validate
+  [yaml path]
+  (let [valid (r/valid? yaml)]
+    (when-not (true? valid)
+      (fail path (::anom/message valid)))))
+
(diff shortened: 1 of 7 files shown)


tool/src/main/fc4/io/render.clj
```diff
@@ -0,0 +1,168 @@
...
+(defn fail
+  ([path msg]
+   (fail path msg {} nil))
+  ([path msg data]
+   (fail path msg data nil))
+  ([path msg data cause]
+   (let [e (if cause
+             (ex-info (err-msg path msg) data cause)
+             (ex-info (err-msg path msg) data))]
+     (if *throw-on-fail* (throw e) e))))
```
arsenerei: I prefer seeing these on separate lines.

tool/src/main/fc4/io/render.clj
```diff
@@ -0,0 +1,168 @@
...
+        :ret  (s/or :valid   nil?
+                    :invalid (partial instance? Exception))
+        :fn   (fn [{{:keys [yaml path]} :args, ret :ret}]
+                (and (= (first yaml) (first ret))
+                     (or (= (first yaml) :valid)
+                         (includes? (.getMessage (second ret)) path)))))
+
+(defn tmp-png-file
+  [path]
+  (-> (file path) (.getName)
```
arsenerei: I'd prefer to see these on separate lines.

tool/src/main/fc4/io/render.clj
```diff
@@ -0,0 +1,168 @@
...
+                    (>= (count (::r/png-bytes result-val)) min-valid-png-size)
+                    (= ret-tag :success)
+
+                    (< (count (::r/png-bytes result-val)) min-valid-png-size)
+                    (and (= ret-tag :failure)
+                         (includes? (.getMessage ret-val) path)))
+
+                  false)))
+
+(defn get-out
```
arsenerei: I'm not certain of the purpose of this function from the docstring.

aviflax: Fixed in 2afd4c6 — thanks!

tool/src/main/fc4/io/render.clj
aviflax: Fixed in 2afd4c6 — thanks!

tool/src/main/fc4/io/render.clj
aviflax: Fixed in 2afd4c6 — thanks!

PR #130, opened 2018-12-17
[tool] Upgrade to Clojure 1.10

aviflax:
Just released today!

PR #132, opened 2018-12-18
Update Clojure in builder image to v1.10

sgerrand:
💁 Missed in #130.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ tool_job_defaults: &tool_job_defaults
   working_directory: ~/project/tool
   docker:
     # The source for this image is at ./images/tool/Dockerfile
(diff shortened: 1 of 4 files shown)


sgerrand:
> I think we should also update the base images of the Dockerfiles used for local testing and packaging

To what?

aviflax:
> To what?

To this new build image hosted at Quay.

E.G.: https://github.com/FundingCircle/fc4-framework/blob/master/tool/Dockerfile.test#L6

The base image used by these local Dockerfiles for local use should match that used in CI.

Seems like it’d be a good idea to add a comment to that effect to the files.

aviflax:
Aha, I see a97ec5f — excellent. That covers `Dockerfile.test` but we need to update `Dockerfile.pkg` as well.

sgerrand:
> worth maybe squashing a few of the commits together as well?

Very much so. Done in 46d53116f29104021d7854bfa12c1d66cdad2aaf. ♻️

PR #134, opened 2018-12-18
Tweaks

aviflax:
Finally a tweaks PR that removes more lines than it adds!


tool/src/main/fc4/integrations/structurizr/express/render.clj
```diff
@@ -47,10 +47,13 @@
...
   []
-  (let [possible-paths [; this first one is used when the tool is packaged in a jar
+  (let [possible-paths [; This first one must be first so it’s preferred to an
+                        ; “installed” executable when running tests from source
+                        "renderer/render.js"
+                        ; Used when the tool is packaged in a jar
                         (str (file (jar-dir) "fc4-render"))
                         "render"
-                        "renderer/render.js"
+                        "target/pkg/renderer/render-mac"
```
sgerrand: 🙋 What's the difference between "mac" and "macos"?

aviflax: I changed the packaging script to use `mac` instead of `macos` to make the package name shorter by 2 chars. Seriously. Because I had just made the package name longer by adding the architecture to it.

PR #143, opened 2019-02-12
[docs] Convert Markdown inside table to HTML

aviflax:
This is a follow-up to #142. I used [grip][grip] to preview the markup wherein Markdown was nested
inside HTML, and it rendered just fine. So that situation renders just
fine within github.com, e.g. when viewing the repo or the readme. I had
thought that GitHub Pages used the same Markdown renderer as github.com,
but apparently not — when this was published to [the main FC4 website][home]
the Markdown markup was escaped, not rendered.

Converting the content in the table to HTML should eliminate this problem.
It’s maybe not super-ideal, but table support in Markdown is iffy in
general, so I can live with this. I think it’s an acceptable situation
until we switch the main website to a custom Jekyll theme and I can
start using CSS to lay out the page rather than tables.

[grip]: https://github.com/joeyespo/grip
[home]: https://fundingcircle.github.io/fc4-framework/

screen shot 2019-02-12 at 12 35 54

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,22 @@ inline CSS. Understandable. So while this is kinda ugly, it more or less works. 
 
 <table><tr><td>
 
-FC4 is a [_Docs as Code_](https://www.writethedocs.org/guide/docs-as-code/) framework that enables software creators to author, publish, and maintain software architecture diagrams more effectively, efficiently, and collaboratively over time.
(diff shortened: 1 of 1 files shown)


aviflax:
Huh, it didn’t work:

screen shot 2019-02-12 at 12 53 52

My mistake for being lazy and not using jekyll to preview locally. I’ll do that now and try another fix.

PR #144, opened 2019-02-12
[docs] HTML fixes for Jekyll compatibility

aviflax:
This is a follow-up to #143, as per [this comment](
https://github.com/FundingCircle/fc4-framework/pull/143#issuecomment-462864842):

> Huh, it didn’t work:
> 
> screen shot 2019-02-12 at 12 53 52
> 
> My mistake for being lazy and not using jekyll to preview locally. I’ll do that now and try another fix.

Apparently for whatever reason Jekyll doesn’t like HTML table tags to be
stacked successively in the same line. Whatever.

PR #145, opened 2019-02-14
Decomplect website and developer docs

aviflax:
…and fix how the home page is rendered too (it’s currently a mess).

Until now, the entire repo (i.e. the root of the repo) has been the
“publication source” for the repo’s GitHub Pages site build. This was
totally fine for awhile, but it’s started to cause issues.

First, I noticed that because I was using `tool/README.md` as both the
README for devs wanting to work on fc4-tool *and* as the source of the
tool page of the website, I had both end-user docs and developer docs on
the same page, which was sub-optimal — it made both kinds of docs less
distinct and less explicit, and has been less than user-friendly.

Then, through the course of multiple PRs (#142, #143, #144) I realized
that the root-level README.md, which was being rendered as the home
page of the repo on github.com *and* as the home page of the website on
GitHub Pages, was being processed by a different Markdown processor for
each context. This made it tricky to figure out what markup would work
where, and meant that any time I’d change any of the README files I’d
need to preview the changes using both grip *and* Jekyll so as to test
how the markup would be translated to HTML in both contexts.

Well, that seemed untenable, and I’d already been planning to decouple
the website (including the end-user docs) from the dev docs (accessible
via the repo at github.com) and move the website into the `docs` dir,
because I’m working on a bunch of new end-user docs and the root dir of
the repo is starting to get really messy in some of those branches.

So there’s a couple of things happening here:

* All end-user docs have been moved into the new root-level dir `docs` which will become the new “publication source” for the website
  * The `md` files under `docs` therefore need be tested (previewed) with Jekyll alone
* The README.md files are now focused on the use case of developers
  viewing them in the context of the repo on github.com — i.e. they’re
  the developer docs
* I added a README.md to `docs` that’s meant to be viewed in the context
  of the repo — this is for developers and/or documentarians who want to
  work on the website

This supersedes #144 — it includes that commit but then effectively overwrites it with a simpler and more comprehensive solution. Given that, I recommend reviewing this PR using the unified diff, rather than reviewing each commit individually. Also, I’m closing #144 because given _this_ work, reviewing that PR would be a waste of time.


aviflax:
😊 Thanks!

Issue #151, opened 2019-03-14
Features to support people using editors that complain when files change


aviflax:
I did a _Getting Started_ micro-workshop with @Ninju last week, and while it went well overall, there was some real awkwardness caused by the `fc4 edit` workflow — his editor didn’t like files that it had open being changed by another process, and it would throw up a bunch of prompts and the default action was to revert to the version in the buffer, rather than to reload from disk. This was annoying.

Sure, @Ninju and anyone using a similar editor could probably change some settings to change their editors’ behaviors, but that’s annoying. It would be better for fc4-tool to somehow accommodate such cases, and help the users out.

Some ideas:

1. Add a new `format` subcommand that would reformat files and then exit — no watching
   1. (unless maybe a `--watch` option was specified).
2. Add a new `--watch` option to the `render` subcommand that would have it stay up and running after the initial render, and then re-render any diagram that’s changed while it’s watching.

I suspect that two enhancements together would address this situation fairly robustly.

@Ninju what do you think?

aviflax:
This was addressed by #153 and #177 — [rendering](https://fundingcircle.github.io/fc4-framework/docs/features#rendering) has been decoupled from [formatting](https://fundingcircle.github.io/fc4-framework/docs/features#formatting) and [snapping](https://fundingcircle.github.io/fc4-framework/docs/features#snapping), so a user can now invoke `fc4` and have it only render diagrams, which means the tool will not modify the YAML files open in the editor at all, in that case.

I suspect that’s sufficient to address this.

PR #153, opened 2019-04-02
[tool] Add new subcommands: format and process

aviflax:
To address, at least to some degree, #151.

@Ninju I’d love to get your 👀 on the new subcommands as described in the updated docs, and also the new descriptions of the tool’s main features in the docs. I’m open to any and all feedback but I’m especially interested to know whether these new commands seem useful to you, and whether they make sense.

One thing I’m a little uneasy about is the name of the command _process_ — it just seems super vague and generic. But I couldn’t think of anything better. Please let me know if you have any ideas!

Thanks!

### DIFF
--- a/docs/tool/index.md
+++ b/docs/tool/index.md
@@ -19,6 +19,36 @@ Methodology](../methodology/):
 > * Renders diagrams
 
 
+## Features
+
+The tool has two main features:
+
+### Formatting
+
+Rewrites [Structurizr Express][structurizr-express] diagram YAML files:
+
+* Removes properties with empty/blank values
+* Removes extraneous quotes around values that are obviously strings
+* Sorts properties and elements using a known and stable sort order
+* “Snaps” the elements and vertices to a virtual grid
+  * i.e. rounds their coordinates
+* Adjusts element coordinates if necessary so as to create top and left margins in the diagram
+  * i.e. ensures that all elements are at least 50px away from the top/left edges of the canvas
+* Offsets the positions of `People` elements so they’ll align properly with other kinds of elements
+  (working around a rendering quirk of Structurizr Express)
+
+
+### Rendering
+
+Given [Structurizr Express][structurizr-express] diagram YAML files, creates PNG image files that
+contain the visualization of the diagram.
+
+* The resulting image file is created in the same directory as the YAML file, with the same filename
+  except its extension is `png`
+  * e.g. `docs/spline_reticulator_01_context.yaml` yields `docs/spline_reticulator_01_context.png`
+* If the image file already exists it will be overwritten
+
+
 ## Setup
 
 ### Requirements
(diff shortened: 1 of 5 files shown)


aviflax:
> Code LGTM 👍

Thank you!

> We should perhaps standardize/generalise the how and what we output to the screen, but otherwise looks good.

Ah, that’s a great point! I’ve pushed 31af58d which makes the output a little more consistent. Not perfect, but an improvement, I think.

> The commands would be useful, yes 🙂 especially `format`.

Great!

> As far as I remember the original problem, it was that I expected `render` to _also_ format the yaml. The `process` command helps here as it's clearer to the end user that `render`'s only job is to `render` the yaml into an image. Therefore, I would say the job's a goodun. 👍

Great, thanks!

> The `process` name, I agree with you, could be better. I have no suggestions.. sorry! 😞

No worries, I still don’t have a better idea either. So I’ll just ship this, and if we ever think of a better name we could always add it as an alias.

> I think it starts to get interesting when we consider the `edit` command.

This is all great feedback, thank you!

> * The `render` part is very slow, making `edit` difficult to use IMO.

I’ve been planning to speed it up, I have a plan — but I wasn’t sure how important it was. It’s helpful to hear that the slowness is indeed a problem for folks beyond myself. (I think it also precludes some useful use cases, such as a CI job that automatically renders all the diagrams in a corpus.)  So I’ll try to get some momentum on the performance improvements I have planned.

> * A user _probably_ doesn't need constant formatting until they're done with all their edits.

Aha, interesting. This highlights for me that “formatting” may be the wrong name/term for that feature, because the feature does more than simply conform the YAML to a certain syntax style guide. It also fixes up the layout of the diagram itself, changing the coordinates of elements and vertices, snapping them all to virtual grids, hopefully aligning things (although that can be made smarter in the future.) (That’s something that can be useful while a person is editing a diagram, because they can move the YAML back and forth between Structurizr Express and their editor, and whenever they hit save, fc4-tool will snap the possibly-messy coords set in Structurizr Express. That’s certainly how I use it.)

🤔 so it _might_ be better to rename that feature to “processing” to explain that it really does those _two_ things. OTOH, perhaps those two things should be decomplected — perhaps they’re really _two_ features that I’ve needlessly complected. Perhaps we should stick with “formatting” to describe the cleaning up of the YAML syntax, and come up with some new name/term for the layout fixups. Maybe “snap”?

Yeah, I’m kinda liking the idea of splitting that feature that’s currently sprawling and hard to describe into two smaller, simpler, more focused features.

If we went down that path, perhaps the tool could have these subcommands:

* `format` — cleans up the YAML syntax so as to facilitate peer review and useful version history
* `snap` — snaps all the coordinates to virtual grids
* `render` — renders to png
* `process` — does all of the above
  * potentially with optional switches to skip/disable the individual features, e.g. `--no-render`
* `edit` — does all of the above, but in a persistent “watch” mode
  * this could potentially be replaced by `process --watch` just to reduce the number of subcommands

What do you think?

> * The `render` part is blocking, meaning I can't continue editing the document and start a new `render` cycle without first waiting for the current one to finish.

Hmm I see. You mean you hit save, the file is immediately formatted and snapped. But then rendering takes awhile — say 10 seconds. During those 10 seconds, if you make changes to the file and hit save, there will be a delay before the file is formatted and snapped. It might be another 5–10 seconds. And if you continue to make changes during that time, I suppose you might then get in trouble because the contents of your buffer could diverge from what’s on disk, and then when fc4-tool does get around to rewriting the file, your editor will have to prompt you to choose one set of changes to keep — either the changes you made in the buffer or the changes made by the tool to the file on disk. Which would be annoying and/or frustrating.

Is that right?

If so, that makes sense to me. This is certainly something that would be partly ameliorated by making rendering faster — but not completely. I think a more comprehensive amelioration might be to have rendering happen in its own queue, separate from formatting and snapping, so that rendering, which is slow, doesn’t block formatting and snapping, which is nearly instant. Does that make sense?

> * There's no need to render a document (or format it) if it's already been formatted and rendered.

I’m guessing you’re thinking about the `render` and `process` commands here, because in the `edit` mode, rendering will only occur when a YAML file is changed, _after_ the tool starts up.

So, yeah, let’s say one runs `fc4 render .` or `fc4 render */**/*.yaml`, effectively instructing the tool to render a large corpus of diagrams. It’d be cool if the tool could determine which YAML files were out of sync with their corresponding PNG files and render *only* those, skipping the ones that would be effectively a no-op. Good idea!

> Perhaps hashing the yaml and using content based addressing would help us in the future here (could be a line in the yaml or a different file, rather than messing with file names which would cause confusion in the git history).

Perhaps! Certainly a promising line of thought.

> Anyway, looks great.

Thanks!

> Thanks for doing this. Should make fc4 a lot easier to use and understand!

My pleasure, that’s great to hear — and thanks for all the feedback and review!

aviflax:
@Ninju I was thinking about the set of possible subcommands I floated above, including `process`, and it occurred to me that we could potentially get rid of `process` and `edit` by moving to a composable switch-based approach rather than a subcommand-based approach.

By “composable switch-based approach” I mean like `tar` which supports various switches that direct its actions, i.e. `tar -x -z -v -f archive.tar.gz` can be shortened to `tar -xzvf archive.tar.gz`

So perhaps `fc4` could have these switches:

* `-f or --format` to format the YAML
* `-s or --snap` to snap the coords to the grid
* `-r or --render` to render the diagrams to PNG files
* `-w or --watch` to stay persistent, watch the supplied paths for changes, and when changes are observed, apply the actions specified via some combination of `-f/s/r`

In this scenario, `fc4 process <path...>` could be replaced by `fc4 -fsr <file>` and `fc4 edit <path...>` could be replaced by `fc4 -fsrw <path…>`. And of course someone could specify just one or two actions, whatever they want.

What do you think?

Issue #155, opened 2019-04-30
Replace Node Renderer with Clojure Renderer


aviflax:
But gradually and safely, of course.

## Background

* When I first added automated rendering I decided to do so by invoking Chrome/Chromium as a headless browser (as a subprocess), loading Structurizr Express (SE) in the browser, and programmatically interacting with SE, which would do the _actual_ rendering
* One of the reasons I chose Chrome/Chromium was because of the existence of Puppeteer, a JavaScript library for programmatically driving Chrome/Chromium via its devtools debugging protocol. Puppeteer provides a powerful API, much of which is way higher-level than the devtools protocol itself.
  * i.e. Puppeteer includes browser automation features that are significantly higher-level than those built into the protocol
* Because I wanted to use Puppeteer, it made sense to me to implement the renderer in a Node program written in JavaScript, such that fc4-tool could invoke and use via RPC (well really just POSIX pipes, but aren’t those RPC?)
* So I did that. And it worked.
* But there were some downsides:
  * The JavaScript code had no unit tests — it was tested indirectly by testing the Clojure code that called the JS code
  * The Renderer could render only one diagram at a time — once for each invocation. This means that every time we want to render a diagram, we have to start Node, launch a browser, and load SE — every single time.
    * i.e. we can’t amortize those time costs over multiple diagrams
  * So [rendering feels slow](https://github.com/FundingCircle/fc4-framework/pull/153#pullrequestreview-225963344)
  * The distribution package is very heavy — it’s 25–27 MB compressed, and when decompressed the renderer binary (`fc4-render` with Node embedded) is 47 MB
  * The CI infrastructure is made more complex by needing to:
    * Have Node installed
    * Download and cache two different sets of dependencies
  * Packaging (and debugging packaging) is way more complex than if this were a Clojure-only project
* Recently I had an epiphany that because Chromium’s devtools debugging protocol is transported via WebSockets, and specified and implemented independently of Puppeteer, there’s no real reason that the renderer couldn’t be implemented in Clojure.
  * So I did some research and I found a Clojure library that seems to provide everything we need to re-implement the renderer in Clojure: [clj-chrome-devtools](https://github.com/tatut/clj-chrome-devtools)

## Plan

1. _Add_ the new Clojure renderer alongside the Node renderer.
   * The Node renderer will remain the default; a user would enable the new Clojure renderer via a new CLI flag, something like `--renderer=experimental`
1. Gather user feedback
[...]

PR #158, opened 2019-04-30
Define Renderer protocol & make Node renderer an implementation of it

### LINKED ISSUE #156: Define Renderer protocol & make Node renderer an implementation of it
Preparatory refactoring to add new Clojure renderer, as described in #155.

See #155 for background.

aviflax:
As preparatory refactoring for adding the new Clojure renderer (#155).

I also simplified and streamlined the value returned when an error occurs.

Closes #156

### DIFF
--- a/tool/deps.edn
+++ b/tool/deps.edn
@@ -1,7 +1,6 @@
 {:paths ["src/main" "src/cli"]
  :deps
  {org.clojure/clojure         {:mvn/version "1.10.0"}
-  org.clojure/data.json       {:mvn/version "0.2.6"}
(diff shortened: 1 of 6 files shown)


aviflax:
Looks like I broke the tests… fixing!

aviflax:
Looks like I fixed the tests — that was more work than I’d thought 😅 .

It took me 4 commits to fix the tests, so this branch is now a little cluttered. I could be convinced to squash it back down to 2–3 commits, either before or after review, but I’m also tempted to leave the 4 fix commits in place, as a permanent record of what happens when you neglect to run the tests locally before pushing…

PR #169, opened 2019-05-20
Update blurb to add “…and documentarians…” to target audience

aviflax:
I’m going to demo the framework in an hour here at Write the Docs
Portland 2019 and I want everyone here to feel included in the target
audience!

This is a little bit of an ambitious expansion of the target audience,
but I feel good about it. Ambitious in a good way; I think it’s (mostly)
ready.


aviflax:
Thanks @alex-steinbaugh! I do, for process reasons. Thanks!

PR #172, opened 2019-06-03
Use Vale to lint prose

aviflax:
I was inspired to give this a try by Jodie Putrino’s excellent unconference talk on Vale at Write the Docs Portland 2019.


sgerrand:
📝 See https://errata-ai.github.io/vale/ for more info.

Issue #168, opened 2019-05-15
Simplify the CLI


aviflax:
In #153, @Ninju and I discussed some issues with the current design of the CLI, especially the different subcommands and what they mean.

Specifically:

* It’s very non-obvious what the `edit` and `process` subcommands do
* The `edit` and `process` subcommands both tightly couple the snapping of coords to a virtual grid, the reformatting of the diagram YAML, and rendering — which doesn’t seem necessary.
* It’d probably be better to just let users choose which of those features they want to be active, on the fly.
* If we use “short options” (e.g. `-r`) rather than subcommands, we could allow them to be grouped (or bundled) together, à la `tar`, e.g. `tar -xzv …`

In that discussion I proposed doing away with the subcommands:

> So perhaps `fc4` could have these switches:
>
> * `-f or --format` to format the YAML
> * `-s or --snap` to snap the coords to the grid
> * `-r or --render` to render the diagrams to PNG files
> * `-w or --watch` to stay persistent, watch the supplied paths for changes, and when changes are observed, apply the actions specified via some combination of `-f/s/r`
>
> In this scenario, `fc4 process <path...>` could be replaced by `fc4 -fsr <file>` and `fc4 edit <path...>` could be replaced by `fc4 -fsrw <path…>`. And of course someone could specify just one or two actions, whatever they want.

I’ve been rolling that over in the back of my head since then, and I feel good about that proposal. So let’s go ahead and do it!

Some notes:

* This will require some preparatory refactoring to decouple snapping from formatting… well worth doing anyway
* This will reduce duplication in the codebase
* This will reduce total LoC
* This will significantly reduce untested LoC

aviflax:
I’ve been thinking that in order to ensure that the design of the CLI is clear, intuitive, and logical, I need to think about future switches/options we might add, so as to make sure we’re using the best set of options as a holistic set.

Some future options I suspect we might want to add, in addition to those specified above:

* `--debug` to activate debug mode
* `-v or --verbose or —verbosity` to toggle/specify verbose mode or maybe a verbosity level
* `-l or --lint` to check formatting
* `-m or --model` to specify the model to use (when the new DSL (#73) is fully implemented+supported)
* `-e or --export` to export FC4 views (as per #73) to Structurizr Express diagrams
* `-o or --output-format` to specify the output format used (png, svg, pdf, etc)
* `--recursive` to toggle whether directories should be searched recursively for diagrams/views

aviflax:
[...]

PR #174, opened 2019-06-17
[tool] Decouple two more features: edit + watch

aviflax:
This is preparatory refactoring for #168.

As part of #168, we’ll be making the watch a mode that’s composable with
the main features, and we’ll be enabling users to toggle each of the
main features individually. In other words, we have three major
features: `-f`, `-s`, and `-r`. A user can invoke the tool with just one of
those — any one, or with any two of them, or with all three. They can
also optionally add `-w` to have the feature(s) applied in “watch mode”.

This is substantially different than the current paradigm, wherein a
specific and limited set of subcommands trigger a hard-coded “workflow”
or mode of operation.

In order to support the new approach, it is necessary to decouple the
watch feature from edit workflow — essentially, to extract the watch
feature from the edit workflow, and make it a higher-order function so
that it can be composed with any workflow (or featureset, mode of
operation, etc.) That’s what this change does.

Most of the code related to watching is moved from `edit.clj` to `watch.clj`
with minimal modifications. The biggest change is that the watch
workflow can no longer assume that if the function it invokes changes
the file that’d just been changed, that that change would be followed by
a rendering step that’d block for a substantial amount of time — at
least 300ms and often 1000–2500. And therefore we can no longer assume
that when processing subsequent events, that the file will still be in
the active set because it’s being rendered. Therefore I restored the map
in the Hawk “context” value that keeps track of when a file change event
was most recently triggered, for each given file. This had been removed
in a5274963eb because at the time it proved sufficient to simply keep
track of a current-active set and filter by membership in that set.

For more background on how the watch has worked to date, and avoided
duplicative work and infinite loops, see a5274963eb.

Anyway, since the tests now supply processing functions that are
extremely fast, and some of them change the subject file, I needed to
restore the time-based successive-event filtering. I kept the active-set
filtering for now, but I’m not sure it’s still needed. This is pretty
hard to reason about, honestly.

### DIFF
--- a/tool/src/main/fc4/io/watch.clj
+++ b/tool/src/main/fc4/io/watch.clj
@@ -1,0 +1,144 @@
+(ns fc4.io.watch
+  "Functions related to watching paths and invoking functions when those paths change."
+  (:require [fc4.io.util :refer [beep debug print-now]]
+            [fc4.io.yaml :refer [yaml-file?]]
+            [hawk.core :as hawk])
+  (:import [java.io File OutputStreamWriter]
+           [java.time LocalTime]
+           [java.time.temporal ChronoUnit]
+           [java.util.concurrent Executors ExecutorService]))
+
+(defn secs-since
+  [inst]
+  (.between (ChronoUnit/SECONDS) inst (LocalTime/now)))
+
+(defn ms-since
+  [inst]
+  (.between (ChronoUnit/MILLIS) inst (LocalTime/now)))
+
+(def ^:private secs-threshold-to-print-event-time 10)
+(def ^:private min-ms-between-changes 1200)
+
+(defn process-fs-event?
+  [{:keys [active-set recent-set] :as _context}
+   {:keys [kind file] :as _event}]
+  (let [in-active-set (contains? @active-set file)
+        in-recent-set (contains? recent-set file)
+        last-processed (get recent-set file)
+        since-processed (when in-recent-set (ms-since last-processed))
+        process? (and (#{:create :modify} kind)
+                      (yaml-file? file)
+                      (not in-active-set)
+                      (or (not in-recent-set)
+                          (>= since-processed min-ms-between-changes)))]
+    (debug (.getName file) "\n"
+           "  in active set:  " in-active-set "\n"
+           "  in recent set:  " in-recent-set "\n"
+           "  last-processed: " (str last-processed) "\n"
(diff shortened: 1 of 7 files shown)


tool/src/main/fc4/io/watch.clj
```diff
@@ -0,0 +1,144 @@
...
+              (fn []
+                ;; I don’t know why, but for some reason when this function is
+                ;; run in the Executor’s thread, *out* appears to be bound to
+                ;; some other writer, rather than the default writer that is the
+                ;; root binding. This re-binds *out* to the writer passed in as
+                ;; out, which enables us to have multiple instances of this
+                ;; workflow run simultaneously and print to independent writers
+                ;; so that the output can be examined in test assertions without
+                ;; concern of those multiple workflows sharing the same writer.
+                (binding [*out* out]
```
alex-steinbaugh: This looked familiar to me...

The middle example illustrates this best:
https://clojure.org/about/concurrent_programming

"[Dynamic Vars] have a (thread-shared) root binding which can be established by def, and can be set using *set!*, _but only if they have been bound to a new storage location thread-locally using binding._"

So it looks like you're doing the right thing setting it this way

With some googling I found another solution that does this with altering the var-root here:
https://blog.raek.se/2011/01/24/executors-in-clojure/

tool/src/main/fc4/io/watch.clj
```diff
@@ -0,0 +1,144 @@
...
+                      (beep) ; good chance the user’s terminal is in the background
+                      (println "🚨" (or (.getMessage e) e)))
+                    (finally
+                      (swap! active-set disj file))))))
+    (assoc-in context [:recent-set file] event-ts)))
+
+(defn start
+  "Starts a hawk watch and returns the watch object, enriched with a few keys
+  specific to this workflow: :executor and :active-set. You can pass the result
+  to stop to stop both the executor and hawk’s background thread, after which
```
alex-steinbaugh: Repetition in "to stop" here.

aviflax: Thanks! I’ll fix this in the follow-up PR.

tool/src/main/fc4/io/watch.clj
aviflax: Thanks! I’ll adjust the comment (I can at least make it shorter) in the follow-up PR.

PR #175, opened 2019-06-18
Tweaks

aviflax:
Yes, really. :trollface: 

I suggest reviewing each commit individually.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,18 @@ jobs:
           key: clj-lint-deps-{{checksum "deps.edn"}}
           paths: [.cpcache, ~/.m2, ~/.gitlibs]
 
+   # Kibit is is a static code analyzer for Clojure. It searches for patterns of code that could be
+   # rewritten with a more idiomatic style. https://github.com/jonase/kibit
+   tool_kibit:
+     <<: *tool_job_defaults
+     steps:
+       - checkout: {path: ~/project}
+       - restore_cache: {keys: ['kibit-deps-{{checksum "deps.edn"}}']}
+       - run: clojure -A:kibit # downloads deps if needed
+       - save_cache:
+          key: kibit-deps-{{checksum "deps.edn"}}
+          paths: [.cpcache, ~/.m2, ~/.gitlibs]
+
    tool_build_dist_pkg:
      <<: *tool_job_defaults
      steps:
@@ -151,6 +163,7 @@ workflows:
       - lint_markdown
       - tool_test
       - tool_clj_lint
+      - tool_kibit
       - tool_build_dist_pkg
       - tool_test_dist_pkg: {requires: [tool_build_dist_pkg]}
       - tool_publish_dist_pkg: {requires: [tool_test, tool_test_dist_pkg]}
--- a/tool/src/main/fc4/model.clj
+++ b/tool/src/main/fc4/model.clj
@@ -1,12 +1,12 @@
 (ns fc4.model
-  (:require [clj-yaml.core           :as yaml]
-            [clojure.set                       :refer [union]]
-            [clojure.spec.alpha      :as s]
-            [clojure.spec.gen.alpha  :as gen]
-            [clojure.string                    :refer [includes? split]]
-            [fc4.files                        :refer [relativize]]
-            [fc4.spec               :as fs]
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -42,6 +42,18 @@ jobs:
...
           paths: [.cpcache, ~/.m2, ~/.gitlibs]
 
+   # Kibit is is a static code analyzer for Clojure. It searches for patterns of code that could be
+   # rewritten with a more idiomatic style. https://github.com/jonase/kibit
+   tool_kibit:
+     <<: *tool_job_defaults
+     steps:
+       - checkout: {path: ~/project}
+       - restore_cache: {keys: ['kibit-deps-{{checksum "deps.edn"}}']}
+       - run: clojure -A:kibit # downloads deps if needed
```
alex-steinbaugh: Gonna steal this from you :)

tool/src/main/fc4/model.clj
```diff
@@ -83,11 +83,10 @@
       ;; however, the target system must be specified, for uniformity. So we
       ;; just add it in right here.
       (update :uses (fn [sys-refs]
-                      (into #{}
-                            (map #(if (:system %)
-                                    %
-                                    (assoc % :system sys-name))
-                                 sys-refs))))
+                      (set (map #(if (:system %)
```
alex-steinbaugh: Feels like `reduce` might be more natural here, i.e.

`(reduce
     #(conj %1 (if (:system %2) %2 (assoc %2 :system sys-name)))
     #{}
     sys-refs)`
 conj should preserve the accumulator type as a set: https://clojure.org/reference/data_structures#Sets

aviflax: Thanks! I’ll take a look.

aviflax: I see what you mean… I think it’s because this is doing _two_ things: (a) adding explicit system names to those system references that don’t already have them, and (b) transforming the sequential collection to a set. And yeah, I think I agree that `reduce` seems like a more elegant way to do those two things together. The thing is, though — maybe those two things shouldn’t be done together! And here’s the thing about `reduce` — maybe this is rigid of me, but I still tend to try to reserve `reduce` for situations wherein the resulting collection will — or at least _might_ — end up smaller than the “input” collection. So I’m not super comfortable with using `reduce` here, because in this case I don’t want to imply that the result might be smaller than the input. The most important work being done here is the selective transformation of the sys-ref maps. So maybe I should rewrite this to split the input collection in two, modify all the maps that meet the criteria (with `map`) and then join the collections back together. And transforming the collection to a set should probably be done discretely of all that.

Does that make any sense?

aviflax: Tried some of this. If I take the `set` out of the equation, that helps.

Then I tried to make the transformation more classically functional.

Here’s what I came up with:

```diff
-      (update :uses (fn [sys-refs]
-                      (map #(if (:system %) % (assoc % :system sys-name))
-                           sys-refs)))
+      (update :uses (fn [refs]
+                      (-> (group-by #(contains? % :system) refs)
+                          (update false (fn [refs] (map #(assoc % :system sys-name) refs)))
+                          (vals)
+                          (->> (reduce concat)))))
```

I guess that’s a little more… I dunno, I guess correct in a way, in that it makes super-duper clear that we’re only transforming a certain subset of the collection that meets a criterion. However, it’s way longer and more verbose… which seems… not great.

Thoughts?

alex-steinbaugh: I see...

To your first point:

> because in this case I don’t want to imply that the result might be smaller than the input

This is a good point, this wouldn't conditionally conj to the accumulator or prematurely terminate w/ `reduced`.  Also, unlike map, reduce reduce isn't lazy which doesn't seem to matter right now but could.

> I guess that’s a little more… I dunno

The first form you had was perfectly readable and a real nitpick on my part to even mention.  But since I have my eye on this namespace, it looks like :tags gets updated twice:

https://github.com/FundingCircle/fc4-framework/blob/3626a01d5a9b47e32eabddceda9a17c1e7b2c118/tool/src/main/fc4/model.clj#L77

Also I think some of the namespace imports got in a weird state formate wise (an even smaller nitpick, but I know we're probably of the same mind on formatting).

PR #176, opened 2019-06-28
Minor cleanup of a few comments and docstrings

aviflax:
As per super helpful insight from @alex-steinbaugh in #174 — thanks Alex!!

### DIFF
--- a/tool/src/main/fc4/io/watch.clj
+++ b/tool/src/main/fc4/io/watch.clj
@@ -68,15 +68,7 @@
   (let [event-ts (LocalTime/now)]
     (.execute executor
               (fn []
-                ;; I don’t know why, but for some reason when this function is
-                ;; run in the Executor’s thread, *out* appears to be bound to
-                ;; some other writer, rather than the default writer that is the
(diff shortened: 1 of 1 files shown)


tool/src/main/fc4/io/watch.clj
```diff
@@ -87,8 +79,8 @@
 (defn start
   "Starts a hawk watch and returns the watch object, enriched with a few keys
   specific to this workflow: :executor and :active-set. You can pass the result
-  to stop to stop both the executor and hawk’s background thread, after which
-  they should be garbage-collected, if you don’t hold on to a reference.
+  to stop both the executor and hawk’s background thread, after which they
```
j-po: it took me a minute to parse what was going on here, but I think the original was meant to read "You can pass the result to `#'stop` to stop both the executor and hawk's background thread...". I think that specificity is important

aviflax: Aha! Yes! Good point! Thanks!

aviflax: Fixed in 571a91a, please re-review, thanks!

PR #177, opened 2019-07-01
Simplify CLI

### LINKED ISSUE #168: Simplify the CLI

aviflax:
I suggest reviewers review each commit individually.

Thank you!

Closes #168


docs/tool/index.md
```diff
@@ -98,44 +98,77 @@ Please see [The Authoring Workflow](../methodology/authoring_workflow.html) sect
...
 
-`fc4 render file ...`
-
-Renders each specified Structurizr Express YAML file as described [above](#rendering)
-
-* One or more file paths must be supplied
-* The resulting image file is created in the same directory as the YAML file, with the same filename
-  except its extension will be `png`
+* The resulting image files are created in the same directory as their corresponding YAML files,
+  with the same filenames except their extensions are `png`
```
j-po: we could say "base filename" (or even "stem") here, but that might be too jargon-y for less dev-y readers. "with the same base filename and the `png` extension` might give enough context clues, though

docs/tool/index.md
```diff
@@ -98,44 +98,77 @@ Please see [The Authoring Workflow](../methodology/authoring_workflow.html) sect
...
+
+If specified, elements in diagrams will be [snapped](#snapping) to a virtual grid.
+
+#### Watching
+
+##### `-w | --watch`
+
+If specified the tool will start up in a persistent mode and watch the diagrams in/under the
+specified paths for changes. Diagrams will be processed (according to the feature options, at least
+one of which is required) when they’re changed, and not when the tool is first invoked.
```
j-po: Maybe something like "If specified, the tool will start up in a persistent mode and watch files in/under the specified paths for changes. When a file is changed, the tool will process it according to the feature options, at least
one of which is required. In this mode, the tool does not process any files when first invoked."

I think "in" rather than "in/under" is enough, but also I may have bought too much into what an old professor of mine used to call "the grand illusion of the directory structure"

tool/src/main/fc4/io/cli/main.clj
```diff
@@ -0,0 +1,181 @@
+(ns fc4.io.cli.main
+  "The CLI command that is the primary interface of the tool."
+  (:gen-class)
+  (:require [clj-yaml.core :refer [parse-string]]
+            [clojure.pprint :refer [pprint]]
+            [clojure.string :as str :refer [join lower-case]]
+            [clojure.tools.cli :refer [parse-opts]]
+            [fc4.io.cli.util :as cu :refer [beep exit fail]]
+            [fc4.io.render :refer [render-diagram-file]]
```
j-po: I know not everyone likes nested imports, but I am definitely a nested imports kiddo and I think they might aid readability here

tool/src/main/fc4/io/cli/main.clj
```diff
@@ -0,0 +1,181 @@
...
+   [nil  "--debug" "For use by developers working on fc4 (the tool)."]])
+
+(def legacy-subcommand->new-equivalent
+  ; This is missing `export` but I’m 99.9% sure that no one was using that. I certainly wasn’t.
+  {"edit"   "fc4 -fsrw"
+   "format" "fc4 -f"
+   "render" "fc4 -r"})
+
+(def strings
+  {:preamble "Usage: fc4 OPTIONS PATH [PATH...]\n\nOptions:\n"
```
j-po: It looks like each of these is only used in one place, so it might be more readable to inline them

tool/src/main/fc4/io/cli/main.clj
aviflax: Hmm, last time this came up, I was convinced by the findability/searchability argument. Is there a rebuttal to that argument?

tool/src/main/fc4/io/cli/main.clj
aviflax: 🤔 I don’t have strong feelings about this… except lately I’ve been on a kick of trying to avoid inlining literals in functions — especially magic numbers and UI copy. I’ve been thinking that it’s probably much better when those things are named and are used by reference. However, I might be erroneously complecting those two cases, magic numbers and UI copy. Maybe it’s way more important/useful with magic numbers… and maybe doing this with UI copy can make the functions that assemble the UI copy together hard to read. OTOH, any application that supports multiple (human) languages in its UI will need to define its strings apart from its logic. So I’ve been thinking that maybe it’s just a “best practice” and a good simple rule of thumb to just always avoid inline literals in functions. I could be convinced either way… do you know of any good write-ups on this?

tool/src/main/fc4/io/cli/main.clj
j-po: I personally find big blocks of imports more legible when they're a bit more structured, but the searchability tradeoff is real. So yeah, no rebuttal beyond my initial comment

tool/src/main/fc4/io/cli/main.clj
j-po: No good write-ups come to mind, unfortunately. I think a big difference between UI copy and magic numbers is that UI copy is self-documenting (outside of the i18n case, where it's only self-documenting in one language). I do think the extent to which keeping UI copy in a place separated from its usage impedes readability is a reasonable tradeoff for i18n, but if we're not doing i18n, I don't think it's necessary

aviflax: Yeah, now that you put it that way, I see that this violates YAGNI. So I’ll change it. Thanks!

tool/src/main/fc4/io/cli/main.clj
```diff
@@ -0,0 +1,181 @@
...
+  If the main document is empty/blank, then the workflow will be short-circuited and this function
+  will return nil without calling f."
+  [yaml-file-contents f]
+  (let [{:keys [::fy/front ::fy/main]} (split-file yaml-file-contents)]
+    (some->> (parse-string main)
+             (f)
+             (stringify)
+             (assemble front))))
+
+(defn- format-and-or-snap
```
j-po: Does this need to be its own function? (Or is completing the split of these two features out of scope for this ticket?)

tool/src/main/fc4/io/cli/main.clj
```diff
@@ -0,0 +1,181 @@
...
+        do-format #(with-msg "formatting" (reformat %))
+        f (cond (and format snap) (comp do-snap do-format)
+                format            do-format
+                snap              do-snap)]
+    (spit file-path
+          (transform-file-contents yaml-file-contents f))))
+
+(defn- process-file
+  [file-path {:keys [format snap render watch] :as options}]
+  ;; Optimization opportunity: this is a little inefficient in that if rendering is specified along
```
j-po: From previous teams, I'm in the practice of signing and dating TODOs and FIXMEs, which gives future readers a bit more context (without having to dig through the `git log` or `git blame`) on how current the messages are and who to talk about if in doubt.

tool/src/main/fc4/io/cli/main.clj
aviflax: Addressed in 4a7ff2c. Thanks!

tool/src/main/fc4/io/cli/main.clj
aviflax: Not really; I only broke it out because `process-file` was too long and harder to follow — in a sense, it was operating at two different depths. With this split, `process-file` seems much easier to follow, to me. That said, I’m certainly open to other suggestions!

tool/src/main/fc4/io/cli/main.clj
aviflax: I’m not entirely sold, but I’m definitely willing to give it a try. Addressed in f06cefd. Thanks!

docs/tool/index.md
aviflax: Addressed in f06cefd. Thanks!

docs/tool/index.md
```diff
@@ -98,44 +98,78 @@ Please see [The Authoring Workflow](../methodology/authoring_workflow.html) sect
...
+
+##### `-s | --snap`
+
+If specified, elements in diagrams will be [snapped](#snapping) to a virtual grid.
+
+#### Watching
+
+##### `-w | --watch`
+
+If specified the tool will start up in a persistent mode and watch the YAML files in/under the
```
j-po: I think this flows better as, "if specified, the tool"

tool/src/main/fc4/io/cli/main.clj
j-po: I didn't phrase my question well. I think it might be clearer to call separate formatting and snapping functions from `process-file`

docs/tool/index.md
aviflax: Agreed, thanks!

tool/src/main/fc4/io/cli/main.clj
aviflax: No worries!

I have to acknowledge that this function is quite awkward and weird. Normally I wouldn’t allow myself to include `and` in a function name, let alone `and` _and_ `or`! Sure, it lets me keep `process-file` high-level, but it also adds a lot of not-fully-necessary indirection and jumping around. (Including `transform-file-contents` which is only used in one place, so its of dubious value as a top-level function.)

I’ve attempted to clean this up in fd0012ac — what do you think?

docs/tool/index.md
aviflax: Addressed via fd0012a — thanks!

docs/tool/index.md
aviflax: Addressed via f06cefd — thanks!

PR #178, opened 2019-07-12
Add CLI option --tmp-renderer

### LINKED ISSUE #159: Add new CLI flag for specifying renderer
Once we have two renderers (#157) we need a way for users to indicate which they want to use.

The option should be `--tmp-renderer` and the supported options should be `experimental` and `stable`. The default, for now, should be `stable`.

The reasons for the `tmp` in the option name is because once we delete the Node renderer (#163) we’ll be back to just a single renderer and the option will be useless. So at that point we’ll just remove it.

Since this is the first end-user-visible change in this project, let’s also update the fc4-tool container diagram as part of this.

* Background in #155
* Blocked by #157 and #168

aviflax:
So as to enable users to opt-in to the experimental (pure Clojure) renderer, as opposed to the default (Node) renderer.

The new renderer itself still has bugs, but that’s expected at this point. In fact, that’s a big motivation for adding this option at this stage: so we can get some help finding bugs.

Closes #159

### DIFF
--- a/tool/src/main/fc4/io/render.clj
+++ b/tool/src/main/fc4/io/render.clj
@@ -10,7 +10,6 @@
             [fc4.files :refer [remove-extension]]
             [fc4.io.util :refer [binary-spit debug err-msg fail read-text-file]]
             [fc4.io.yaml :as yaml]
-            [fc4.integrations.structurizr.express.node-renderer :refer [->NodeRenderer]]
             [fc4.rendering :as r :refer [render]]
             [fc4.spec :as fs])
   (:import [java.io File FileNotFoundException]))
@@ -98,13 +97,12 @@
   Express diagram definition, rendering it to an image, and writing the image to
   a file in the same directory as the YAML file. Returns the path of the PNG
   file that was written (as a string) or throws an Exception."
-  [in-path]
-  (with-open [renderer (->NodeRenderer)]
-    (let [yaml     (read-text-file in-path)
-          _        (yaml/validate yaml in-path)
-          result   (render renderer yaml)
-          _        (debug (::anom/message result))
-          _        (check result in-path)
-          out-path (yaml-path->png-path in-path)]
-      (binary-spit out-path (::r/png-bytes result))
-      out-path)))
+  [in-path renderer]
+  (let [yaml     (read-text-file in-path)
+        _        (yaml/validate yaml in-path)
+        result   (render renderer yaml)
+        _        (debug (::anom/message result))
+        _        (check result in-path)
+        out-path (yaml-path->png-path in-path)]
+    (binary-spit out-path (::r/png-bytes result))
+    out-path))
--- a/tool/test/fc4/io/cli/main_test.clj
+++ b/tool/test/fc4/io/cli/main_test.clj
@@ -4,7 +4,6 @@
             [clojure.string :refer [includes? split-lines upper-case]]
             [clojure.test :refer [deftest is testing]]
             [clojure.tools.cli :refer [parse-opts]]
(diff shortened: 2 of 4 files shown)


tool/src/main/fc4/io/render.clj
```diff
@@ -98,13 +97,12 @@
...
-      out-path)))
+  [in-path renderer]
+  (let [yaml     (read-text-file in-path)
+        _        (yaml/validate yaml in-path)
+        result   (render renderer yaml)
+        _        (debug (::anom/message result))
+        _        (check result in-path)
+        out-path (yaml-path->png-path in-path)]
+    (binary-spit out-path (::r/png-bytes result))
+    out-path))
```
aviflax: This hunk _looks_ like a lot was changed but in fact the change is fairly minor: the `with-open` form was removed (no longer needed because the renderer is now passed in) and its contents were outdented accordingly but not otherwise changed.

tool/test/fc4/io/render_test.clj
```diff
@@ -26,7 +27,7 @@
   (System/setProperty "apple.awt.UIElement" "true")
   (require '[image-resizer.core :refer [resize]]))
 
-(reset! iou/debug? true)
+(reset! iou/debug? false)
```
j-po: I'm curious as to why this was changed, and more broadly whether a default value for it needs to be written in here at all

tool/test/fc4/io/cli/main_test.clj
```diff
@@ -141,37 +140,67 @@
...
+                            (main/-main "-fsr" "--tmp-renderer=experimental" (str yaml-fp)))))
+              _ (is (.canRead (file actual-png-path)))
+              difference (->> (map binary-slurp [expected-png-path actual-png-path])
+                              (map bytes->buffered-image)
+                              (map #(resize % 1000 1000))
+                              (reduce image-diff))]
+          (is (not= yaml-file-size-before (.length yaml-fp)))
+          (is (= (main-doc yaml-expected) (main-doc yaml-fp)))
+          (is (<= difference max-allowable-image-difference))
+          (is (= 4 (count-substring output "✅")) output)
```
j-po: I think in general it would be nice if we could test some of these things in a way other than parsing the string output for signs of success, but 1) we do I guess also need to test the string output somehow, and 2) that's out of scope for this ticket

tool/test/fc4/io/render_test.clj
aviflax: > I'm curious as to why this was changed…

🤔 honestly I don’t remember exactly why… but I have a feeling that it was because I realized I want the tests to be “quiet” by default. I was running tests frequently while working on this, and I _think_ maybe the debug output was getting mixed in with the non-debug output and making it hard for me to validate (with my eyes, I guess) the output of the tool in various scenarios.

> …and more broadly whether a default value for it needs to be written in here at all

Yeah, it’s a very good point. This honestly does feel like a smell. Because it’s not actually a default, it’s actually more of a state change. It feels like this state change shouldn’t really be here, in this specific test namespace, because it’s a global change that impacts any and all behavior going forward. What happens when someone else working on a different test namespace wants `debug?` to be `true` whenever _those_ tests are run? So yeah, this feels like it should probably be higher level, at the test-runner or test suite level.

I’ll add a comment to that effect before I merge.

Thanks!!!

tool/test/fc4/io/cli/main_test.clj
aviflax: Makes sense. Feel free to create a ticket to keep track of these ideas!

PR #181, opened 2019-07-16
Reproduce and fix bug in Clojure renderer

### LINKED ISSUE #180: Clojure renderer is hanging or failing with large and large-ish diagram YAML files
In some cases, it hangs indefinitely. In some others, it returns correctly and _seems_ to have succeeded, but the PNG file that it returns is a visualization of the nearly-empty default diagram that is the default zero state of Structurizr Express (inline below).

Blocks #161

----

Here’s that zero-state PNG file:

structurizr-express

aviflax:
Wherein it would hang when rendering any diagram wherein the output PNG
was more than 1MB in data-size.

Fixes #180


j-po:
Looks good to me. It looks like the fix was to set `:max-msg-size-mb` to a larger value, then you also added some logging changes to help you find the issue and a timeout so that similar errors don't hang the app in the future. Is that about right?

aviflax:
@j-po yes, exactly right!

tool/src/main/fc4/util.clj
```diff
@@ -112,3 +113,16 @@
   [message]
   {::anom/category ::anom/fault
    ::anom/message message})
+
+(defmacro with-timeout
+  "If the timeout elapses before the body has completed, a TimeoutException will be thrown with a
+  not-particularly-helpful message."
+  ;; To consider: perhaps we should include part of the body in the exception message?
+  {:derived-from "https://stackoverflow.com/q/6694530/7012"}
```
j-po: Is there tooling that uses this metadata?

aviflax: Nope, I made it up, ad-hoc.

Issue #163, opened 2019-05-01
Remove (delete) the Node renderer and all related infrastructure, etc.


aviflax:
… including the CLI option `--tmp-renderer`

* Background in #155
* Blocked by #162

Issue #69, opened 2018-09-06
Tool tests are way slow


aviflax:
In Funding Circle’s internal Slack, @AndreaCrotti asked why the tests run so slowly.

I replied:

> The tests run slowly mainly because they are mostly generative tests, i.e. property tests, using generators mostly auto-generated via clojure.spec. Some of the functions operate on fairly large data structures and so these fairly large structures are being generated thousands of times for each run of the test suite.
> 
> At least, that’s my hypothesis. I haven’t confirmed it. I have tried tuning the tests, though; those that operate on the largest data structures are performing hundreds of test iterations rather than the default of 1000.
> 
> e.g. https://github.com/FundingCircle/fc4-framework/blob/master/tool/test/fc4/integrations/structurizr/express/edit_test.clj#L11
> 
> This is a known problem that I have on my radar to address.

aviflax:
This has gotten worse recently; up from 6–8 minutes to 10–11. Getting more pressing.

I don’t think it’s related to the recent upgrade to Clojure 1.10 (#130 + #132) because as per [the CircleCI Insights page](https://circleci.com/build-insights/gh/FundingCircle/fc4-framework/master) the slowdown seems to have started about 5 days ago:

<img width="1431" alt="screen shot 2018-12-18 at 10 11 06" src="https://user-images.githubusercontent.com/141844/50162961-79593d00-02ad-11e9-97e3-b7bc84ece4d3.png">

aviflax:
* Running the tests is doubly slow when measuring coverage, which is the default setup for the CI job
* To speed up the CI test runs we might want to try running tests [with more cores](https://circleci.com/docs/2.0/configuration-reference/#resource_class) or [on multiple containers in parallel](https://circleci.com/docs/2.0/parallelism-faster-jobs/).

aviflax:
Update:

* I recently got a new laptop:
  * It’s a MacBookPro15,1 with an Intel i7-8850H CPU @ 2.60GHz, 16 GB 2400 MHz DDR4 RAM, and a fast SSD. It’s running MacOS 10.14.5 (18F203) and Java openjdk version "11.0.4" 2019-07-16 (AdoptOpenJDK (build 11.0.4+11))
  * The full test suite on master as of this moment (bc5542cbf7962ed5692941bac8733fcf168ef062), _without_ coverage measurement, takes 3m 41s
[...]

PR #183, opened 2019-07-24
Add performance tests for renderers

### LINKED ISSUE #182: Add performance tests for renderers
My initial motivation is to ensure that the new Clojure renderer (#155) is faster than the current Node renderer. Once we decommission the Node renderer (#163) these tests will serve as regression tests.

aviflax:
My initial motivation is to ensure that the new Clojure renderer (#155)
is faster than the current Node renderer. Once we decommission the Node
renderer (#163) these tests will serve as regression tests.

Closes #182

### DIFF
--- a/tool/test/fc4/integrations/structurizr/express/chromium_renderer_test.clj
+++ b/tool/test/fc4/integrations/structurizr/express/chromium_renderer_test.clj
@@ -149,6 +149,14 @@
               (is (s/valid? ::r/failure-result result)
                   (expound-str ::r/failure-result result))
               (is (every? #(includes? message %) expected-strings))
-              (is (includes? message "errors were found in the diagram definition")))))))))
+              (is (includes? message "errors were found in the diagram definition")))))))
+    (testing "performance"
+      (let [yaml (slurp (file dir "diagram_valid_formatted_snapped.yaml"))
+            start-ns (System/nanoTime)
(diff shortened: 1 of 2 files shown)


j-po:
I agree with the need for performance regression tests, but adding 25 seconds to every test run is not amazing. I'd prefer if we could add a way to run all but these tests (e.g. by entry point using `circleci.test`-style selectors)

aviflax:
> I agree with the need for performance regression tests, but adding 25 seconds to every test run seems like a little much. I'd prefer if we could add a way to run all but these tests during normal development (e.g. by entry point using `circleci.test`-style selectors)

I agree with this, but I’m not sure the scope of this branch+PR should be expanded to account for this. We _already_ have a known test suite duration problem (e.g. #69). Sure, it’s not great to make it worse, but it really only makes the suite a little longer than it already is.

So, how would you feel about us working on improving the test suite duration outside of the context of this branch/pr? I’d rather we ship this and deal with the problem being even more acute — perhaps it will finally spur action on that front.

aviflax:
To be more specific:  I just ran the test suite on my system on this branch and on master, and excluding JVM startup and code compilation, the suite durations were:

* master: 207.788 seconds
* this branch: 226.956 seconds

So that’s a difference of 19.168 seconds; ~9% longer.

j-po:
Sure, that makes sense :shipit:

Issue #179, opened 2019-07-12
Add file name to progress output of tool when passed multiple files and invoked without watch mode


aviflax:
In watch mode, the tool always outputs the name of the changed file just before processing it.

Currently, outside of watch mode, it does not. So there’s no way to tell which files have errors:

```
$ fc4 -fs *.yaml 
reading+parsing+validating...✅ formatting...✅ snapping...✅
reading+parsing+validating...✅ formatting...✅ snapping...✅
reading+parsing+validating...✅ formatting...🚨class java.lang.Character cannot be cast to class java.util.Map$Entry (java.lang.Character and java.util.Map$Entry are in module java.base of loader 'bootstrap')
reading+parsing+validating...✅ formatting...✅ snapping...✅
```

So let’s output the file names in both modes — immediate mode and watch mode. It’s a good thing anyway, all things being equal, for the output of the tool to be as consistent as possible, across modes.

aviflax:
What about something like this?

```diff
diff --git a/tool/src/main/fc4/io/cli/main.clj b/tool/src/main/fc4/io/cli/main.clj
index 2170369..69d71c6 100644
--- a/tool/src/main/fc4/io/cli/main.clj
+++ b/tool/src/main/fc4/io/cli/main.clj
@@ -101,14 +101,14 @@
          result#)))
 
 (defn- process-file
-  [file-path renderer {:keys [format snap render watch] :as options}]
+  [file-path preamble renderer {:keys [format snap render watch] :as options}]
   ;; Optimization opportunity: this is a little inefficient in that if rendering is specified along
   ;; with formatting and/or snapping then we’re reading the YAML file twice. So we should probably
   ;; refactor render-diagram-file to accept diagram-yaml rather than a file-path. (It accepts a
   ;; file-path for historical reasons, so that the old edit workflow could call it directly. Since
   ;; we’ve removed the old edit workflow, we can change it.) (Avi Flax, July 2019)
   (try
-    (print-now "reading+parsing+validating...")
+    (print-now preamble " → reading+parsing+validating...")
     (let [yaml-file-contents (read-text-file file-path)]
       ;; Optimization opportunity: the YAML is parsed by both validate and below if format and/or
       ;; snap are true. (Avi Flax, July 2019)
@@ -150,10 +150,9 @@
 (defn- start
   [renderer {paths                       :arguments
              {:keys [watch] :as options} :options}]
-  (let [f #(process-file % renderer options)]
-    (if watch
[...]

PR #184, opened 2019-07-27
add filename printing to non-watch runs

j-po:
addresses #179

### DIFF
--- a/tool/src/main/fc4/io/cli/main.clj
+++ b/tool/src/main/fc4/io/cli/main.clj
@@ -154,8 +154,7 @@
     (block (watch/start #(process-file % renderer options) paths))
     (doseq [p paths]
       (print-now p)
-      (process-file p renderer options))
-    (run! f paths)))
+      (process-file p renderer options))))
 
 (defn -main
   [& args]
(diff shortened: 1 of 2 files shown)


j-po:
- Sorry for not clearly marking this as WIP. tests are upcoming
- I generally try to avoid the indirection of tersely-named intermediary bindings like `f`, but I'll try to come up with something that meets both of our needs
- I was considering a solution inside of the file processing function, and I'm happy to revisit that

aviflax:
No worries, sounds good!

j-po:
one wrinkle in using the same codepath for both is that a lot of the logic in the existing printing logic for watch only really makes sense in that context.

aviflax:
> one wrinkle in using the same codepath for both is that a lot of the logic in the existing printing logic for watch only really makes sense in that context.

Yeah, I thought that might be the case. I’ll take a look and see if I can think of something, but don’t let this slow you down; it’s hasn’t met the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) threshold for duplication so let’s not prematurely apply [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) — do what seems best to you.

j-po:
debugging an issue where nothing prints in watch mode. I think for some reason `process-file` isn't being run

aviflax:
This work has stalled as @j-po and I are no longer on the same team. I’m going to take it over, but I’m closing this PR for now and will probably open a new PR as a follow-up soon.

PR #188, opened 2019-08-01
Change recommended installation method to Homebrew

aviflax:
We just published a [Homebrew Tap][1] for Funding Circle’s FLOSS tools and [that tap][2] includes a formula for fc4-tool.

This is great because not only does it make it _way_ easier to install the fc4 tool, it also makes it _much_ more likely that users will install updates!

(This _might_ also enable us to get ahold of some download/install stats, but that might require contributing a formula to homebrew-core; I’m not sure.)

[1]: https://docs.brew.sh/Taps
[2]: https://github.com/FundingCircle/homebrew-floss

### DIFF
--- a/docs/tool/index.md
+++ b/docs/tool/index.md
@@ -53,6 +53,22 @@ contain the visualization of the diagram.
 
 ## Setup
 
+### Quick start for Homebrew users
+
+If you already use [Homebrew][homebrew], one or both of these commands might be  all you need to get
+started:
+
+```shell
+# If you’re using MacOS and you don’t already have Chromium or Chrome installed:
+brew cask install chromium
+# If you’re using a different OS and you don’t already have Chromium or Chrome installed
+# then install Chromium or Chrome however you generally install such software on your system.
+
+# The main event (should work on any OS that supports Homebrew)
+brew install fundingcircle/floss/fc4
+```
+
+
 ### Requirements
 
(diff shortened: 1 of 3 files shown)


aviflax:
@miketheman thanks for the review! I’ve just got to adjust the branch protection settings for this repo and then I’ll merge.

I’d love to hear how you came across the framework, whether you’ve used it or not, and any feedback whatsoever that you’d care to share. (Either here or in a new issue, whatever you prefer.) Thank you!

docs/tool/index.md
```diff
@@ -53,6 +53,22 @@ contain the visualization of the diagram.
 
 ## Setup
 
+### Quick start for Homebrew users
+
+If you already use [Homebrew][homebrew], one or both of these commands might be  all you need to get
+started:
+
+```shell
+# If you’re using MacOS and you don’t already have Chromium or Chrome installed:
```
sgerrand: 💅 Apple tweaked the name (yet again) a year or so ago.
```suggestion
# If you’re using macOS and you don’t already have Chromium or Chrome installed:
```

aviflax: Ah yes, I was vaguely aware — but I don’t like it. It’s a twee affectation. It’s a proper noun and should be capitalized. Thanks for the suggestion though, I appreciate the attention to detail!

PR #196, opened 2019-08-27
Use CircleCI Docker images for testing and packaging

aviflax:
… instead of our own custom image.

Thanks for getting the ball rolling on this, @sgerrand!


sgerrand:
Thanks for following up on my initial commit and finishing it off. 🏅

PR #197, opened 2019-08-29
Make the Clojure renderer the default

### LINKED ISSUE #161: Make the Clojure renderer the default
- Background in #155
- Blocked by #160

aviflax:
Closes #161


aviflax:
Thanks Alex!

aviflax:
I agree about the case statements, but anyway they’ll be going away soon when I delete the Node renderer (#155)

PR #208, opened 2019-09-18
Update deps and CI environment

aviflax:
At first I thought this was going to be simply bumping a few version
numbers, which I do occasionally as a matter of course.

Then I recalled that [a change][1] had recently been made to
clj-chrome-devtools that would enable it to be included as a gitlib
rather than a Maven artifact. I much prefer gitlibs, so I switched it.
Well, that lead to a cascade of changes.

First I ran into [this bug in tools.deps][2] that broke pretty much
everything — but only in CI, not locally. It was pointed out to me in
Clojurians that that bug had already been fixed, so I figured that I
must be using an out-of-date version of Clojure in CI. Sure enough, the
CircleCI Clojure images are fairly out of date and don’t include that
bugfix.

I worked around that by switching from CircleCI’s Clojure images to
CircleCI’s OpenJDK images, and installing the latest (as of this
writing) version of Clojure as a step in the jobs that require run the
project’s code (as opposed to e.g. linting-type jobs that only analyze
the code).

So then the tests passed, but packaging was failing. Turns out I ran
into [another bug related to gitlibs][3], this time in the packaging
tool we’re using, Cambada. I was able to work around this bug by
switching to a fork of Cambada, which seems like a good move anyway
because the “root” Cambada repo seems to be unmaintained.

I _think_ that explains everything, but please let me know if I missed
anything.

[1]: https://github.com/tatut/clj-chrome-devtools/issues/21#issuecomment-520040669
[2]: https://clojure.atlassian.net/browse/TDEPS-52
[3]: https://github.com/luchiniatwork/cambada/issues/9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,20 +7,18 @@ version: 2
 # steps *except* `checkout` to be a subdirectory of the project — because by
 # default, `checkout` checks out into the specified `working_directory`. I found
 # this solution here: https://stackoverflow.com/a/50570581/7012
-tool_job_defaults: &tool_job_defaults
-  working_directory: ~/project/tool
-  docker:
-      # See https://circleci.com/docs/2.0/circleci-images/
-    - image: circleci/clojure:openjdk-11-tools-deps-browsers
 
 jobs:
    tool_test:
-     <<: *tool_job_defaults
+     docker:    # See https://circleci.com/docs/2.0/circleci-images/
+       - image: circleci/openjdk:11-stretch-browsers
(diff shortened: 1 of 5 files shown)


aviflax:
Bonus: these updates seem to have resulting in a significant speedup in running the tests while measuring coverage.

On my laptop, doing so takes ~330 seconds on master but only ~196 seconds on this branch.

Not bad!

(I also checked for a difference in running the tests _without_ measuring coverage and I didn’t see any difference in that case.)

PR #216, opened 2019-10-04
Add a CI job that lints with Eastwood

aviflax:
I think I might have a problem. A linter problem.

Note to reviewers: although it looks like there were a lot of lines changed in this PR, and that’s technically true, most of those are simply moving a bunch of spec definitions from two files to two other files, so that Eastwood wouldn’t freak out about namespaces and filenames not matching. I was planning to make that change anyway, so this seems like a fine time to do so. Anyway, my point is: there’s no need to scrutinize the lines that add specs to `model.clj` or `view.clj`.


.circleci/config.yml
```diff
@@ -58,6 +58,20 @@ jobs:
           key: kibit-deps-v1-{{checksum "deps.edn"}}
           paths: [.cpcache, ~/.m2, ~/.gitlibs]
 
+   # Eastwood is Yet Another lint tool for Clojure.
+   tool_eastwood:
+     docker:    # See https://circleci.com/docs/2.0/circleci-images/
```
sgerrand: 💭 Not sure why this comment is needed. Perhaps you could explain the reasoning for adding it.

.circleci/config.yml
```diff
@@ -58,6 +58,20 @@ jobs:
...
+       - image: circleci/clojure:tools-deps
+     working_directory: ~/project/tool
+     steps:
+       - checkout: {path: ~/project}
+       - restore_cache: {keys: ['eastwood-deps-v1-{{checksum "deps.edn"}}']}
+       - run: bin/install-clojure
+       - run: clojure -A:eastwood # downloads deps if needed
```
sgerrand: Suggest adding this comment in a description for the step instead.
```suggestion
       - run:
           command: clojure -A:eastwood
           name: Download Clojure dependencies, if needed
```

tool/src/main/fc4/image_utils.clj
```diff
@@ -17,7 +17,8 @@
 (defn buffered-image->bytes
   [^BufferedImage img]
   (let [baos (ByteArrayOutputStream.)]
-    (ImageIO/write img "png" baos)
+    (or (ImageIO/write img "png" baos)
+        (throw (IOException. "No appropriate writer could be found.")))
```
sgerrand: 👏

tool/src/main/fc4/model.clj
```diff
@@ -6,9 +6,143 @@
...
+                                 ; Really just trying to influence the
+                                 ; cardinality of the generated value... this
+                                 ; might be a silly way to do it — should
+                                 ; probably use a more explicit approach.
+                                 :min-count 2 :max-count 2)))))
+
+(s/def ::user
+  ;; ::uses is required because in FC4 there’s no point in describing a user
+  ;; unless they *use* one or more systems.
+  ;; TODO: should probably use a different variant of ::uses
```
sgerrand: 💭 Suggest adding an issue on the project instead of an inline "TODO".

.circleci/config.yml
aviflax: Well, not all Docker images have clear documentation available. I’ll clarify the comment. Thanks!

.circleci/config.yml
aviflax: I see what you mean, I can do something like this. Downloading the dependencies is a side effect of the step, but the name of the step would be a good place to note that. Thanks!

tool/src/main/fc4/model.clj
aviflax: Thanks! I’m not going to move this right now, as this isn’t actually new as of this PR; it’s simply been moved as-is from one file to another. But I agree with the point; I’ve created issue #217: “Move all TODOs in the code to GitHub Issues” — thanks!

.circleci/config.yml
aviflax: Addressed in 3c39516 — thanks!

.circleci/config.yml
aviflax: Addressed in 3c39516 — thanks!

aviflax:
I’m merging this despite the drop in coverage reported by codecov, because that drop is caused primarily by the addition of a single new line of code; a line of code that’s both difficult to test and not particularly important; it’s mainly a formality because Eastwood was failing because the code was ignoring the result of a call to a Java method that returns a value. I think that value being false is extremely unlikely so I’m not going to worry about testing it right now.

Issue #236, opened 2019-12-03
Compilation error: 'fc4.integrations.structurizr.express.chromium-renderer' not found


mathpunk:
I'm attempting to get up and running on an Arch Linux system, and I cannot get past this error. I've tried loading this namespace in the repl, and I've tried running tests. I'm getting more or less the same error but not more information (that I understand) about how to continue.

mathpunk:
Note: I have chromium at version 78 and google-chrome-stable at 77. That could cause problems but I don't know if it's causing this problem.

aviflax:
Hi @mathpunk thanks so much for the bug report! I really appreciate it. I’ve never used Arch before, but now that I have an Arch user wanting to use FC4, I’d be delighted to try to add it to the set of test environments. I’ll look into this and report back within a day or two. Thanks again!

aviflax:
Ah, I just re-read your bug report and it registered that you’ve already tried to do some debugging, so obviously you’re familiar with Clojure to some degree. I just want to check: is your goal to simply _use_ FC4, or are you also interested in working _on_ FC4?

Thanks!

mathpunk:
Ooh, well I am mostly interested in using it, but the reason I got excited to use it is because there is some great work happening in the mathematics of diagram languages so, maybe in the future I could contribute.

aviflax:
Very cool! I’m intrigued by the work you mentioned; if you have a chance please do share some links!

(So far I’ve stubbornly avoided employing any sort of mathematics or graph theory into FC4 but that’s just due to my own biases and imposter syndrome, etc. It’s been clear to me for awhile that the brute-force relationship traversal in the code right now will need to be replaced with graph traversal algorithms over graph data structures, e.g. using [Loom](https://github.com/aysylu/loom) or something similar.)

mathpunk:
Hmmmm. I checked out `bug-236`. When I ran `clj -A:dev:test` and then 

```
(do
     (require '[eftest.runner :refer [find-tests run-tests]])
     (run-tests (find-tests "test") {:fail-fast? true}))
```
I got

```
Syntax error (FileNotFoundException) compiling at (fc4/integrations/structurizr/express/chromium_renderer.clj:1:1).
Could not locate clj_chrome_devtools/automation__init.class, clj_chrome_devtools/automation.clj or clj_chrome_devtools/automation.cljc on classpath. Please check that namespaces with dashes use underscores in the Clojure file name.
``` 
[...]

PR #237, opened 2019-12-04
Misc Tweaks

aviflax:
Yes, really 😬.

(The changes to the CircleCI config are partly preparatory refactoring for addressing #236.)


.circleci/config.yml
```diff
@@ -37,51 +37,60 @@ jobs:
...
+       - run:
+          name: Run cljfmt
+          command: clojure -A:lint
 
    # Kibit is is a static code analyzer for Clojure. It searches for patterns of code that could be
    # rewritten with a more idiomatic style. https://github.com/jonase/kibit
    tool_kibit:
      docker:
-       - image: circleci/clojure:tools-deps
+       - image: circleci/openjdk:11
```
AndreaCrotti: 💅 but you can make this a reference so you only define the docker image once

mattsoutherden:
> save the dependency cache right after downloading/installing dependencies, rather than at the end
of a job.

👍 

This also has the addidtional benefit of avoiding polluting the cache with build/test artefacts that could have unexpected effects on subsequent runs.

.circleci/config.yml
aviflax: Yes, good point! Come to think of it, there’s a lot of repetition in the config that can be consolidated using more recent CircleCI features. I’ll try to do something about that soon-ish. 😬

PR #238, opened 2019-12-04
Add support for Arch Linux

### LINKED ISSUE #236: Compilation error: 'fc4.integrations.structurizr.express.chromium-renderer' not found

aviflax:
…to fix bug #236.

@mathpunk I couldn’t assign this PR to you, I guess because you’re not part of the @FundingCircle org, but your feedback on these changes would be most welcome! You can download and test a built installation artifact from [this pre-release](https://github.com/FundingCircle/fc4-framework/releases/tag/prerelease_bug-236).

## TO DO BEFORE MERGING

1. [x] Merge #237
2. [x] Rebase onto master
3. [x] Re-target this PR to master

----

Fixes #236
Blocked by #237

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,26 +8,37 @@ version: 2.1
 # default, `checkout` checks out into the specified `working_directory`. I found
 # this solution here: https://stackoverflow.com/a/50570581/7012
 
-# Many of the jobs use CircleCI’s docker images, such as circleci/openjdk or circleci/clojure.
+# Some of these executors use docker images provided, maintained, and supported by CircleCI.
 # Those images are documented here: https://circleci.com/docs/2.0/circleci-images/
+
+executors:
+  circleci-openjdk-11:
+    docker:
+      # This image is based on Debian GNU/Linux 9 (stretch)
+      # https://hub.docker.com/layers/circleci/openjdk/11-browsers/images/sha256-c5b820f7ff08cff4f0695ac63461391f570eea0eba0be202ca88e9ecc0c4ef9a
+      - image: circleci/openjdk:11@sha256:c5b820f7ff08cff4f0695ac63461391f570eea0eba0be202ca88e9ecc0c4ef9a
+  circleci-openjdk-11-browsers:
+    docker:
+      # This image is based on Debian GNU/Linux 9 (stretch)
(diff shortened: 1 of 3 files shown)


tool/bin/install-sys-deps
```diff
@@ -0,0 +1,29 @@
...
+
+case "$OS" in
+'linux-arch')
+  # We’re using sudo in /bin/install-clojure
+  pacman --refresh --sync --needed --noconfirm --noprogressbar sudo jre11-openjdk-headless chromium
+
+  # shellcheck disable=SC1091
+  source /etc/profile.d/jre.sh
+  ;;
+'linux-debian')
```
AndreaCrotti: do we need the debian/mac cases if they are doing nothing?

aviflax: Well, I included them because I want the script to fail if a non-supported OS is specified, or if no arg is supplied. Is there a different approach you’d suggest?

.circleci/config.yml
```diff
@@ -206,10 +218,17 @@ workflows:
     jobs:
       - lint_prose
       - lint_markdown
-      - tool_test
+      - tool_test:
+          name: tool_test_debian
+          docker-image: circleci/openjdk:11-stretch-browsers
```
mattsoutherden: I'm guessing it's implicit that the openjdk image is build on debian? What happens if it changes?

aviflax: Yes, it is.

I suppose we should use the image ID (hash) here so the image can’t change on us. I’ll do that. And I’ll add a comment that it’s Debian. Thanks!

aviflax: Done in 25a865a (with some follow-up in 8d44dfc) — thanks Matt!

tool/bin/install-sys-deps
aviflax: As you pointed out in chat, the error message would have been confusing if someone specified an incorrect argument. I’ve attempted to fix this in e2d53c9, although I did introduce a bug. I’ll open a tiny follow-up PR with the fix to the fix. 🙃

Thanks!

PR #239, opened 2019-12-05
Render SVG output to .svg files

### LINKED ISSUE #222: SVG images don’t render in GitHub Pull Requests
In #200 we added support for rendering diagrams to SVG images instead of, or in addition to (simultaneously with) PNG images.

When I implemented that, I had the tool save the SVG images as files with an `.html` extension, because that was what Structurizr Express (SE) did. I didn’t know why that was the case, but I trusted that there was a good reason for that choice, so I figured I’d do the same thing. After all, when I opened those HTML files — both those exported directly from SE and those created by fc4 — in a browser — they rendered nicely. So it seemed good to me.

Unfortunately, I neglected to account for an important goal of FC4: to enable peer review of changes to diagrams via GitHub Pull Requests, via textual diffs of the YAML sources and via visual diffs of the rendered images. GitHub will render SVG files with the `.svg` extension (as long as their contents are valid) but it won’t render any files with the `.html` extension.

This is bad. So let’s fix the tool to save the SVG image files as valid SVG with the `.svg` filename extension.

aviflax:
## Notes to reviewers:

* I suggest you review each commit individually; there’s at least one commit here that’s preparatory refactoring.
* Don’t let that massive LoC++ number intimidate you; it’s mostly SVG files that are either examples or used in the tests.

## TO DO BEFORE MERGING

1. [x] Merge #238
2. [x] Re-point to master
3. [x] Rebase on master

----

Fixes #222
Blocked by #238


aviflax:
@miketheman @99-not-out  this might be relevant to your interests.

tool/src/main/fc4/integrations/structurizr/express/renderer.clj
```diff
@@ -323,10 +293,13 @@
...
     (str test-data-dir (get filenames it "????"))
     (slurp it)
-    (time (render @renderer it))
-    (or (::r/png-bytes it)
-        (println (or (::anom/message it) "WTF")))
-    (when it (binary-spit "/tmp/diagram.png" it)))
+    (time (render @renderer it {:output-formats #{:svg :png}}))
+    (do
+      (if-let [image (get-in it [:fc4.rendering/images :fc4.rendering/svg :fc4.rendering.svg/conjoined])]
+        (spit "/tmp/diagram.svg" image)
```
AndreaCrotti: not sure if it matters but this would not work on Windows probably.
And even on *nix it might be better to use something like this to get a temporary directory:
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory%28java.nio.file.Path,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...%29

tool/src/main/fc4/integrations/structurizr/express/renderer/png.clj
```diff
@@ -0,0 +1,38 @@
...
+        w (max (width di) (width sk))
+        divider-height 2
+        gap 1
+        key-title-y-offset 0 ; Currently 0 for test-compatibility with prior renderer, but I plan to increase this to ~40.
+        key-title-x-offset 35 ; Mainly for test-compatibility with prior renderer, but looks OK.
+        ky (+ (.getHeight di) gap)
+        kx (- (/ w 2) (/ (width sk) 2))
+        h (+ ky (height sk))
+        ci (BufferedImage. w h BufferedImage/TYPE_INT_RGB)]
+    (doto (.createGraphics ci)
```
AndreaCrotti: quill or dali or some other svg library would save you lots of interop here

tool/src/main/fc4/integrations/structurizr/express/renderer.clj
aviflax: Ah, good catch! However this is in a `comment` form so it’s really for developers, not users. If one day some hypothetical developer working on FC4 is using Windows, then they’d be welcome to change this. So I’ll leave it as-is for now. Thanks!

tool/src/main/fc4/integrations/structurizr/express/renderer/png.clj
aviflax: Agreed. It could also potentially unify the code paths for SVG and PNG (i.e. vector and raster) because e.g. Dali has the ability to render a “document” to SVG (XML) _or_ PNG.

However, this isn’t new code; in this PR it’s just being moved, so I’m not going to change it as part of this PR.

Thank you!

simonbrowndotje:
The unnecessary &nbsp; entities were being introduced by the JavaScript rendering framework being used, and have now been removed from the exported SVG. SVGs should now render in GitHub; e.g. https://gist.github.com/simonbrowndotje/3eeeb2283f8c3c70995ec9c6931d02ff

aviflax:
@simonbrowndotje excellent — thank you!

Issue #226, opened 2019-11-05
Proposal: redefine this project as a tool or toolset rather than a “framework”


aviflax:
Today, FC4 is a:

> …framework that enables software creators and documentarians to author, publish, and maintain software architecture diagrams more effectively, efficiently, and collaboratively over time.

I’ve been calling it a _framework_ because for a good while I was just as focused on its methodology as its tooling. As I described in [the blog post introducing the project](https://engineering.fundingcircle.com/blog/2018/09/07/the-fc4-framework/), I asked my friend @ilangoodman if he could think of a singular term for a thing comprised of a tool and a methodology, he suggested “framework”. It made sense to me so I went with it.

Over time, however, I’ve noticed that the terms “framework” and “methodology“ are doing the project no favors. They seem to be vague and off-putting, and they seem to be dampening interest in the framework. On the other hand, when I’ve tried describing FC4 as a *tool* the response has been uniformly positive and eager. People seem to gravitate towards tools and be wary of more amorphous things such as frameworks.

I’ve come to believe that we should call FC4 a tool, and we should focus on, and emphasize, its tool-ness.

It can and will still have a methodology, but instead of being _the_ methodology, it would be _a_ methodology for using the tool — essentially a set of recommended practices, or guidelines, or something along those lines. This would all be part of the docs for the tool. ([The website](https://fundingcircle.github.io/fc4-framework/) would be reorganized accordingly.)

I’d love some feedback on this proposal. Anyone have any thoughts?

/cc @miketheman @ifyoumakeit @sldblog

sldblog:
👍 I am in favour of using "tool" for the reasons you mentioned.

In the [Legal Aid Agency architecture diagrams][laa-diagrams], we use the phrase

> We use the [FC4 toolset][fc4-toolset] to normalise the source of C4 diagrams authored with [Structurizr Express][se].

[laa-diagrams]: https://github.com/ministryofjustice/laa-architectural-diagrams
[se]: https://structurizr.com/help/express
[fc4-toolset]: https://fundingcircle.github.io/fc4-framework/methodology/toolset.html

aviflax:
Thanks Dávid!

miketheman:
Thanks for writing this - it helps clarify the intent of this particular tool.

[...]

PR #242, opened 2019-12-09
Fix a CI script

aviflax:
As per commit comments.


tool/bin/install-sys-deps
```diff
@@ -6,13 +6,13 @@
 
 set -eux
 
-OS="$1"
-
-if [ "$OS" == '' ]; then
+if [ "$*" == '' ]; then
```
mattsoutherden: This could be `if [ -z "$*" ]; then`

aviflax: Thanks! I couldn’t remember that operator. I’ll make this change. Thanks!

.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
        - restore_cache:
            keys:
              - test-deps-v1-{{checksum "deps.edn"}}-{{checksum "bin/download-test-deps"}}
-       - run: bin/install-sys-deps << parameters.os >>
+       - run: ../.circleci/bin/install-sys-deps << parameters.os >>
```
miketheman: If this file is colocated with the config file, do you need to traverse down to step back up, or could you use 'bin/install...' instead?

miketheman: Oh, you were doing that already. So let me ask: why did you need to traverse directories? Add a comment to make it clear.

aviflax: Good point; this did in fact trip me up while working on this fix. The `..` traversal is due to the `working_directory` property of the job, which is just above the job’s steps, being set to `~/project/tool`.

I’ll add a comment to that effect before merging.

This is yet another reason to simplify the repo structure while implementing #226.

aviflax:
> Since you're already renaming the file, would you consider adding a file extension like '.sh' for clarity?

I’m certainly open to it! Your question spurred me to try to do a little research on the pros and cons of using such an extension for shell scripts. For example, I read all the answers to [this question](https://unix.stackexchange.com/questions/31760/file-extensions-for-unix-shell-scripts). There are a _lot_ of opinions about this!

For me this boils down to the utility of being able to see at a glance that a script is a _shell_ script, vs the utility of being able to change the language of a script (e.g. to Python, Joker, etc) without having to change the name of the file and therefore changing its name everywhere it’s called. Right now I’m on the fence on this, so I’m going to sleep on it and see if my brain figures it out overnight

Thanks for the review!

miketheman:
I found further support for no extension in the Google Style Guide https://google.github.io/styleguide/shell.xml#File_Extensions

aviflax:
Good find Mike! Thanks for taking the time to look into this more. I think I’m inclined to leave the filename of the script as-is, without an extension. Can you live with that?

tool/bin/install-sys-deps
aviflax: Done in 58ae250 — thanks!

miketheman:
Yep, no extension makes sense. I am going to think about that style guide more for other projects, since the idea of an executable (binary, even though bash isn't a binary...) without an extension feels more natural to the user.

mattsoutherden:
I concur. 👍

aviflax:
OK, great. Now just waiting for someone to give me that sweet, sweet: ✅ before I merge.

Issue #245, opened 2019-12-17
Support for Bash 3 was accidentally broken in #244


aviflax:
#244 restored support for Java 8 by making the `fc4` launcher script “smarter”. Unfortunately I made it too smart; I used a Bash 4/5 feature that doesn’t work in Bash 3. I only discovered this because Bash 3.2 is the default shell on MacOS, and the homebrew-floss repo CI jobs run on MacOS and therefore caught the error, [for example](https://app.circleci.com/jobs/github/FundingCircle/homebrew-floss/54).

So: we need to restore support for Bash 3 of course, but I also want to add a CI job or two to _this_ repo that’ll catch this sort of thing sooner.

PR #246, opened 2019-12-17
Fix bug #245

aviflax:
## Notes to reviewers

* I recommend reviewing each commit individually


.circleci/bin/install-run-deps
```diff
@@ -0,0 +1,28 @@
...
+
+case $OS in
+'linux-arch')
+  pacman --refresh --sync --needed --noconfirm --noprogressbar jre11-openjdk-headless chromium
+
+  # shellcheck disable=SC1091
+  source /etc/profile.d/jre.sh
+  ;;
+'linux-debian')
+  # Nothing to do here right now, because the Docker image that we’re using for this OS already has
```
mattsoutherden: Is it worth explicitly specifying the deps here? Even if the package manager decides there is nothing to do, it gives you the benefit of codifying exactly what the deps are. How much would is add to the execution time for apt to run?

.circleci/bin/install-test-deps
```diff
@@ -0,0 +1,29 @@
...
+set -eux -o pipefail
+
+OS=$(~/project/.circleci/bin/which-os)
+
+case $OS in
+'linux-arch')
+  pacman --refresh --sync --needed --noconfirm --noprogressbar jre11-openjdk-headless chromium
+  ;;
+'linux-debian')
+  # Nothing to do here right now, because the Docker image that we’re using for this OS already has
```
mattsoutherden: Same as above.

.circleci/bin/install-run-deps
aviflax: Good question!

In this specific case, a call to `apt-get install openjdk-11-jre-headless` would indeed be a no-op and therefore fast, but it would only cover Java, not Chrome/Chromium (Chrome is added to the Docker image by the scripts that CircleCI uses to build their images.) Since it’d be incomplete, I don’t think it’d be worth doing.

What do you see as the core value proposition of your idea? Is it documentation? Because I could think of a better place to document this. In fact, I did add (as part of this PR) a README to the same dir, that tries to document clearly which system dependencies are required in which scenarios. Or is it, rather, perhaps a sort of enhanced resiliency in the CI jobs? The idea being that if, say, someone changed the Docker image in which some of the jobs ran, and in one or more cases they inadvertently used a Docker image that didn’t have, say, Java, or a browser, that then maybe rather than failing the CI job would still work? Or… something else?

mattsoutherden: Both. 😄 

Mostly it was the enhanced resiliancy of explicitly requesting all the deps to be installed. But also to a lesser degree the docs.

Not a blocker either way. Just an idea.

aviflax: Got it, thanks! I’ll take a look at enhancing both.

> Not a blocker either way. Just an idea.

You say that, but where’s my ✅ , huh?

You :trollface:ing me?

mattsoutherden: ~Absolutely!~

No. Just that Mark had already approved it.

aviflax: What do you think of 9ec9f05?

> No. Just that Mark had already approved it.

Ah, I see. New team expectation calibration in progress!

.circleci/bin/install-test-deps
aviflax: Discussed above.

.circleci/bin/install-run-deps
aviflax: I suspect 9ec9f05 will look OK to you, so I’m going to merge… but I’d still like to hear your feedback!

PR #248, opened 2019-12-19
Simplify shell scripts that install sys deps in CI

aviflax:
In aaa945430828 we merged #246, which introduced a whole mess of new
shell scripts for installing system dependencies needed by our CI jobs.

Unfortunately I tried to make those shell scripts “smart” and
inadvertently made them way too complex. They’ve been nagging at me so
I’ve played around with a few different approaches to simplifying them.

The approach I ended up with is so simple as to be simplistic, but I
think that’s probably for the best.

## Notes to reviewers

* I suggest reviewing the diff of `config.xml` first

## Before merging

* [x] #247 must be merged
* [x] re-point this pr to master
* [x] rebase onto master (after review is complete)


mattsoutherden:
> I suggest reviewing the diff of config.xml first

For a minute, I had a horrible vision that you'd replaced the shell scripts with some sort of makefile-like xml. 🙀

aviflax:
> For a minute, I had a horrible vision that you'd replaced the shell scripts with some sort of makefile-like xml. 🙀

Well, actually… I did spend the better part of a day going down a path something like that. Except EDN rather than XML, and without the dependency-tree computations… take a look if you’d like, I posted it up in [a gist](https://gist.github.com/aviflax/8645341be315d7f85ad415fe3614f277). I got to that point and then I realized it was getting too complicated, and it wasn’t even done. Basically it wasn’t worth it. That’s when I switched to the simplistic approach seen here in this PR.

Thanks for the review!

mattsoutherden:
While I can't say that the CircleCI YAML DSL makes me jump for joy, I agree that, given the current requirements, it's the best option.

PR #252, opened 2019-12-30
Update tests to handle changes to Structurizr Express

aviflax:
I was working on something unrelated and the rendering tests started
failing with output like this:

```
expected: (<= distance-percentage (:svg max-allowable-image-differences))
  actual: (not (<= 0.052500776638707676 0.05))
```

as you can see, the rendered SVG image was 5.25% different than the
expected image, which exceeded the max allowable difference of 5%.

I took a look and it seems that Simon Brown has made some minor changes
to Structurizr Express’ rendering; mostly to the watermark that
Structurizr Express has been adding to rendered images. Well, actually
he seems to have removed the watermark and replaced it with a larger,
more opaque Structurizr wordmark in a fixed location.

This is of course 100% fine; it’s his product and it’s his prerogative
to change it however he sees fit, at any time. I have zero problems with
him making this change unannounced; when I chose to have FC4 “wrap” SE
and invoke its rendering routines programmatically, I knew that there
were zero guarantees that SE would continue to function, or to be
available, and I knew that it would be my responsibility to update FC4
if and when SE changed or was retired.

So! There are two changes here:

1. Update the “expected” image to one rendered with the new wordmark
2. Increase the maximum allowable image difference to 6%

I made the latter change because the differences in the rendering here
were so visually trivial, to my eye, that it wasn’t necessary for the
tests to fail in this case.

## Notes to Reviewers

1. I think you can skip over the diff of the SVG file — it’s mostly just noise. Random IDs, that sort of thing.

### DIFF
--- a/tool/test/data/structurizr/express/diagram_valid_expected.svg
+++ b/tool/test/data/structurizr/express/diagram_valid_expected.svg
@@ -2,91 +2,91 @@
 <svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 1748 1740">
   <g>
     <g xmlns:xlink="http://www.w3.org/1999/xlink" id="v-3" class="joint-viewport" transform="matrix(1,0,0,1,0,0)">
-      <g id="j_12" model-id="d12c5ebd-c853-48a1-8e49-d0b06692d618" class="joint-theme-default joint-cell joint-type-structurizr joint-type-structurizr-diagrammetadata joint-element" data-type="structurizr.diagramMetadata" transform="translate(20,1224)">
-        <g id="v-93">
-          <text class="structurizrDiagramMetadata structurizrMetadata" id="v-94" font-weight="normal" font-size="22px" text-anchor="start" fill="#777777" pointer-events="none" font-family="Open Sans"/>
+      <g id="j_12" model-id="16c12989-7fd2-44de-9a18-c8de6a52f73b" class="joint-theme-default joint-cell joint-type-structurizr joint-type-structurizr-diagrammetadata joint-element" data-type="structurizr.diagramMetadata" transform="translate(20,1224)">
+        <g id="v-91">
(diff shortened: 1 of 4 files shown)


aviflax:
> LGTM.

Thanks Alex!

> The testing for this has all been on a MacBook so far, yes? I can test this on some other machines w/ different display resolution later today to double check.

Well, sort of… the tests are fairly comprehensive and they do run on multiple operating systems in our CI environment (Debian, Arch, and MacOS) — but I believe those are all headless “machines.”

Anyway, I’m not sure what display resolution would have to do with the changes in this PR? Maybe it’s not clear how the rendering tests work? Basically, I have a “reference” file (or maybe a few files) that contains the rendered diagram, so that’s effectively the “expected” output of any rendering done in the tests. While the tests are running, they render a diagram source to an image (either PNG or SVG, or both) and then compare the results (as either bytes, in the case of PNG, or as strings, in the case of SVG) to the reference/expected file. The two images being compared don’t have to be _perfectly_ identical; just _nearly_ identical. So I use these “diff” routines to compute the differences between the images, as a percentage, and then check that those differences don’t exceed some semi-arbitrary maximum that I set at some point.

Therefore, I don’t _think_ display resolution should be a factor here, since we’re rendering data to images, on headless machines, and then comparing the images _as data_.

Does that help? Or did I miss something?

alex-steinbaugh:
@aviflax I think I understand, I assumed you were doing some manual inspection here.  This is more about the change in watermark than the underlying SVG image.

I still approve!

aviflax:
Great, thank you!!!

simonbrowndotje:
A couple of notes about the watermarking:

- Watermarking only happens for anonymous users on the cloud service, authenticated users with free accounts on the cloud service, and free on-premises installations ... plus there's a random element to the appearance of the watermark.
- Watermarking does *not* appear with any of the paid subscriptions (both cloud service and on-premises installations).

PR #254, opened 2019-12-30
Simplify directory structure

aviflax:
As per #226, we’re redefining FC4 as a tool rather than a framework.

Until now, FC4 has been conceived of as a framework comprising a tool
and a methodology. The repo therefore contained the tool in the dir
`tool` and the methodology in `docs/methodology`.

Going forward, it’s simply a tool; the methodology will become the user
manual for the tool.

Therefore we can and should simplify the structure of the repo/project;
we should move the contents of `tool` out into the root (or wherever is
appropriate) and then delete `tool`. And so I have.

This is the first step of that work. It’s not complete, but will be
followed quickly by some follow-up changes:

* Refactor the “methodology” docset as the user manual
* Refactor the contents of docs/tool/README.md
  * Basically atomize the contents and move them around to appropriate
    places
* Add redirects for the website pages that have been moved
* Reduce the number of files/dirs in the root of the project
* Lots of copy changes to rename and reframe the project
* etc

## Before Merging

1. [x] #252 must be merged
1. [x] re-point this PR to master
1. [x] rebase this branch onto master

----

Refs #226


aviflax:
Thanks Alex! For both the review and your thoughts. I’m looking forward to working more closely together!

PR #255, opened 2019-12-31
Fix flaky test on MacOS in CircleCI

aviflax:
This test was intermittently failing with output like this:

```
a body that takes longer than the timeout to finish
  expected: (thrown? TimeoutException
                     (u/with-timeout 100 (do (Thread/sleep 150) true)))
  actual: nil
```

And sometimes the inverse.

I assume this has something to do with the virtualization setup, or the fact that the MacOS executors on CircleCI are seemingly slow in general.

### DIFF
--- a/test/fc4/util_test.clj
+++ b/test/fc4/util_test.clj
@@ -58,8 +58,8 @@
 (deftest with-timeout
   (testing "a body that finishes executing before the timeout has elapsed"
     (is (true? (u/with-timeout 100
-                 (do (Thread/sleep 50) true)))))
+                 (do (Thread/sleep 10) true)))))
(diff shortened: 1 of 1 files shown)


mattsoutherden:
IIUC this uses futures, so the caller and callee are running in separate threads. As such, you're trying to keep the margin our error greater than that of the thread scheduler, is that correct?

aviflax:
> IIUC this uses futures, so the caller and callee are running in separate threads.

Yeah, the impl of `with-timeout` uses `future`, so yes.

> As such, you're trying to keep the margin our error greater than that of the thread scheduler, is that correct?

Yeah, I think that’s right. A good way to sum this up. Thanks!

PR #263, opened 2020-01-07
Attempt to fix Codecov

aviflax:
I’m still having trouble with Codecov seemingly ignoring this config file. I did some research and these changes *might* help. They aren’t helping with the GitHub “status checks” on *this* commit while in a “feature branch” — but it’s possible that Codecov really only uses the config file that’s in the default branch? I don’t know, I’m baffled and aggravated with Codecov. In a different branch I’m trying out some other services like Coveralls and Code Climate.

### DIFF
--- a/.circleci/codecov.yml
+++ b/.circleci/codecov.yml
@@ -3,6 +3,8 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 1 # up to 1% drop should pass
-        base: auto
-    patch: no
+        threshold: 1%
+    # I have trouble making sense of the patch reports.
+    patch: off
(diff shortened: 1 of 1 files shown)


mattsoutherden:
There must be a better way to test this kind of thing than having to blindly flail about with commits to master on a live product. (That's not a jibe at you, more a general thought on the functioning of this sort of tool)

aviflax:
I agree 100%. I’m _really_ tired of Codecov. I’m hoping to replace it with Coveralls or Code Climate or something similar sometime soon(ish).

miketheman:
I'll lend my vote for Code Climate, as it can provide more than only coverage.

PR #267, opened 2020-01-09
Fix: CI job publish_dist_pkg fails on master

aviflax:
This is a follow-up to #266 / 99f481bc1b. In that commit I modified the
CI job publish_dist_pkg to always run on master. Unfortunately I
neglected to do the same for its dependencies: build_dist_pkg and
test_dist_pkg. So publish_dist_pkg did indeed run on master after that
PR was merged, but it failed, because the distribution package hadn’t
been created. Doh!


aviflax:
@mattsoutherden perhaps we should simply never halt when on master? i.e. always run all jobs on master… would probably be safer, simpler, clearer, and easier to debug and troubleshoot. Also more concise… I could build it right into the shell script and then we wouldn’t need this `command` entity in the config file. What do you think?

mattsoutherden:
That sounds like a sensible approach, and also a safety net.

PR #271, opened 2020-01-16
Fix long tests redux

aviflax:
## Notes to reviewers

  * Except maybe the first two. The second commit reverts the first, so they sum up to a no-op. I’ve only left them in the branch for historical purposes. Someone searching the commit history at some future point _might_ find those commit messages helpful. Feel free to skip or skim these first two commits.
* Thank you!

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -6,7 +6,7 @@
  ;; “überjar” for this tool (see the :uberjar alias below) can’t access the installation deps.edn
  ;; and doesn’t want to make assumptions for us. For more info see:
  ;;   https://github.com/luchiniatwork/cambada/issues/19#issuecomment-490013695
- :mvn/repos {"central" {:url "http://central.maven.org/maven2/"}
+ :mvn/repos {"central" {:url "https://repo1.maven.org/maven2/"}
              "clojars" {:url "https://repo.clojars.org/"}}
 
  :deps
--- a/src/fc4/dsl/model.clj
+++ b/src/fc4/dsl/model.clj
@@ -24,10 +24,10 @@
 (s/def ::m/tags
   (s/map-of ::m/tag-name (s/or :boolean boolean?
                                :string  ::fs/short-non-blank-simple-str
-                               :strings (s/coll-of ::fs/short-non-blank-simple-str :gen-max 10)
+                               :strings (s/coll-of ::fs/short-non-blank-simple-str :gen-max 2)
                                :number  number?
-                               :numbers (s/coll-of number? :gen-max 10))
-            :gen-max 5))
+                               :numbers (s/coll-of number? :gen-max 2))
+            :gen-max 2))
 
 (defn- url?
   [v]
@@ -42,7 +42,7 @@
        (.isAbsolute (URI. s))))
 
 (s/def ::m/links
-  (s/map-of ::fs/short-non-blank-simple-str (s/and url? absolute-url?) :gen-max 5))
+  (s/map-of ::fs/short-non-blank-simple-str (s/and url? absolute-url?) :gen-max 2))
 
 (s/def ::m/name ::fs/short-non-blank-simple-str)
 
@@ -51,7 +51,7 @@
 (s/def ::m/refs
   (s/coll-of ::m/ref
              :distinct true
-             :gen-max 10))
+             :gen-max 2))
 
 (s/def ::m/system ::m/ref)
 (s/def ::m/container ::m/ref)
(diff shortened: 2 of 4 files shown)


mattsoutherden:
> Anyway, one of our test jobs (Test & Measure Coverage (Debian, Java 11)) went from 8m 18s on master to 3m 59s on this branch. So this clearly
makes a difference.

Does it repeatably make such an improvement? Does the case of 3 for gen-max always not make an improvement?

aviflax:
> Does it repeatably make such an improvement?

Good question — I’ll test this more rigorously.

> Does the case of 3 for gen-max always not make an improvement?

Hmm well, in general the larger the value the more computation required. Does that answer your question? I’m not sure I understood it properly.

aviflax:
> Good question — I’ll test this more rigorously.

Here we go:

First I checked out the current HEAD of master: 2b7ad3ea4.

Then I modified `test/fc4/test_runner/runner.clj`:

```diff
--- a/test/fc4/test_runner/runner.clj
+++ b/test/fc4/test_runner/runner.clj
@@ -6,7 +6,7 @@
             [eftest.report.junit    :as ju]
             [eftest.runner          :as runner :refer [find-tests]]))

-(def test-dir "test")
+(def test-dir "test/fc4/dsl")

 (def output-path
   "This is optimized for CircleCI: https://circleci.com/docs/2.0/configuration-reference/#store_test_results"
@@ -39,6 +39,9 @@
      ;; https://github.com/weavejester/eftest/#multithreading
      :multithread? :vars

+     ;; reproduce the CircleCI bug with java.lang.Runtime/availableProcessors
+     :threads 36
+
      ;; We have lots of tests that take too damn long.
      :test-warn-time 10000

```

Then I started a Docker container using the same base image that we use for our CI jobs (using [this modified version](https://github.com/FundingCircle/fc4-framework/compare/tweaks#diff-115240426f0726b1acf91e4cbd0c69d4) of `bin/docker-test-run` which made it possible for me to check out different commits without restarting the container). The container was limited to 2 vCPUs just like our CircleCI test containers.

Then I ran this:

```shell
$ for run in {1..10}; do clojure -A:test:test/run; done
```

I then checked out 5e5cc10fd and repeated these tests, and then did the same on a1fcc5814. The only
difference was that before checking out a1fcc5814 I removed the `:threads 36` addition to
`runner.clj`, because that commit adds its own value for that key (that’s the point of that commit).

The results:

| commit | times (wall-clock seconds) | median |
| :-- | :-- | :-- |
| 2b7ad3ea4 | 97 162 232 150 147 235 207 203 168 171 | 169
| 5e5cc10fd | 11.8 14.4 15.9 12.4 9.3 12.9 10.8 14.8 10.3 12.9 | 12.65
| a1fcc5814 | 16.1 14.3 14.1 13.9 12.0 21.8 14.3 11.1 13.3 14.6 | 14

I think this shows fairly compellingly that 5e5cc10fd makes a substantial difference for these two tests. Running merely these two tests, however, fails to demonstrate an improvement from a1fcc5814. This makes sense to me, because the problem that a1fcc5814 attempts to address is that when we are running _more than two_ computationally intensive tests concurrently, our performance will suffer, because on CircleCI we’re currently using 36 threads, which should lead to some degree of CPU contention between those threads, because our CircleCI containers have only 2 vCPUs — the more tests we run concurrently, the more contention.

Therefore I ran the tests again on 5e5cc10fd and a1fcc5814, but this time _all_ the tests, and again
when I ran 5e5cc10fd I hard-coded `:threads 36` again so as to simulate running on CircleCI.

This time I ran the test suite 5 times on each commit because running _all_ the tests takes awhile
no matter what, and I was running short on time:

| commit | times (wall-clock seconds) | median |
| :-- | :-- | :-- |
| 5e5cc10fd | 259 246 240 256 242 | 246
| a1fcc5814 | 251 253 255 227 215 | 251

I think it’s pretty clear that a1fcc5814 doesn’t do anything for us at this point. Honestly, that’s
not shocking; I crafted that change on a *different* branch, one that adds a third
computationally-intensive test to the same namespace. In *that* case, trying to run *3* such tests
concurrently on *3* threads *does* have a negative impact when running on a “machine” with only two
“CPUs.” I think I added it on to *this* branch because it seemed like it would pair nicely with
5e5cc10fd — grouping commits by theme is a weakness of mine.

That said, it’s more important to only ship commits that have a concrete positive impact *now* so
I’ve removed a1fcc5814 from this branch and I’ll be including it instead in my next PR.

Thanks for the prompt to quantify the impact of these changes!

Issue #23, opened 2018-05-04
Spooky STDOUT warnings when running `wcb`


damien:
From my Clojure repl:

```
user=> (wcb)
nil
user=> WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by clojure.lang.Reflector (file:/Users/damien.wilson/.m2/repository/org/clojure/clojure/1.9.0/clojure-1.9.0.jar) to method sun.awt.datatransfer.SunClipboard.isDataFlavorAvailable(java.awt.datatransfer.DataFlavor)
WARNING: Please consider reporting this to the maintainers of clojure.lang.Reflector
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

PR #279, opened 2020-02-06
Fix specs related to loading models

aviflax:
…which was merged earlier this week via #23.

I was reviewing some of this code while resolving some merge conflicts resulting from rebasing a local branch onto master, and I noticed that a few things were off in some of the specs in `src/fc4/dsl/model.clj` and `src/fc4/io/dsl.clj`.

## Notes to reviewers

* Thank you!

### DIFF
--- a/src/fc4/dsl/model.clj
+++ b/src/fc4/dsl/model.clj
@@ -223,11 +223,15 @@
     (catch ParserException e
       (fault (str "YAML could not be parsed: error " e)))))
 
+;; NB: this is used in at least one other namespace.
+(s/def ::parse-file-result
+  (s/or :invalid-or-malformed  ::anom/anomaly
+        :valid-and-well-formed map?))
+
 (s/fdef parse-file
   :args (s/cat :v (s/alt :valid-and-well-formed ::file-yaml-string
                          :invalid-or-malformed  string?))
-  :ret  (s/or :valid-and-well-formed ::file
-              :invalid-or-malformed  ::anom/anomaly)
+  :ret  ::parse-file-result
   :fn   (fn [{{arg :v} :args, ret :ret}]
           (= (first arg) (first ret))))
 
@@ -246,15 +250,11 @@
 
 (s/def ::err-msg string?)
 
-;; Note: this is used in at least one other namespace.
-(s/def ::parse-file-result
-  (s/or :valid   nil?
-        :invalid ::err-msg))
-
 (s/fdef validate-parsed-file
   :args (s/cat :parsed (s/alt :valid   ::file
                               :invalid (s/and map? #(not (s/valid? ::file %)))))
-  :ret  ::parse-file-result
+  :ret  (s/or :valid   nil?
+              :invalid ::err-msg)
   :fn   (fn [{{arg :parsed} :args, ret :ret}]
           (= (first arg) (first ret))))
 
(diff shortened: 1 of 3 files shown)


src/fc4/dsl/model.clj
```diff
@@ -223,11 +223,15 @@
     (catch ParserException e
       (fault (str "YAML could not be parsed: error " e)))))
 
+;; NB: this is used in at least one other namespace.
```
AndreaCrotti: Just to understand, why do you need this NB?

test/fc4/io/dsl_test.clj
```diff
@@ -1,70 +1,102 @@
...
             [fc4.util             :as u])
   (:import [java.io FileNotFoundException]))
 
 (u/namespaces '[fc4 :as f]
               '[fc4.model :as m])
 
+;; At first I just called instrument with no args, but I ran into trouble with some specs/fns inside
+;; clj-chrome-devtools. So I came up with this overwrought approach to instrumenting only the functions in
+;; the namespace under test.
+(->> (ns-interns 'fc4.io.dsl)
```
AndreaCrotti: Probably for a separete PR but If you use Kaocha as test runner you don't need any of this, it does the instrumentation for you.
In general I think that instrumenting in each test file never quite works, since if you work with others someone will forget and some specs will not be checked.

src/fc4/dsl/model.clj
aviflax: Well, specs are often defined in a namespace and then used only in that namespace — i.e. by code in the same file. In this case, we have a spec in a different file that’s using this one, so I wanted to warn future maintainers so they’d be aware of this uncommon dependency before changing or removing this spec.

AndreaCrotti: Well that should be caught by tests anyway, and I guess they would do a simple search before removing a spec?
Anyway we can leave the NB I was just curious to know why.

test/fc4/io/dsl_test.clj
aviflax: Thanks, I agree. I have a comment somewhere else about moving this stuff to the project’s test-runner-runner. (One of the reasons I haven’t done so is because I’m also considering just deleting the test-runner-runner and using Kaocha instead.) (I’m stalling on Kaocha because it doesn’t support parallelism when running tests, and this test suite is heavy on computation due to the heavy usage of generative property testing, so I _think_ it benefits substantially from the parallelism support in eftest.)

AndreaCrotti: Parallelism can still be achieved easily by marking tests in different ways (unit/integration etc) with different jobs in CircleCI.
Is not exactly the same thing but you get the same result in the end imho.

src/fc4/dsl/model.clj
aviflax: 👍 🙏

test/fc4/io/dsl_test.clj
aviflax: That makes sense for CI, and seems like a decent solution (if a bit contrived) but it wouldn’t help when running the entire test suite locally… although perhaps that’s an edge case? I suppose I do that less and less frequently these days; I more commonly run a specific test namespace(s) for the namespace(s) I’m working on, then once _those_ pass I’ll push to GH so CI will run the full test suite for me… 🤔

aviflax: Thanks for spurring my thinking along!

PR #281, opened 2020-02-11
Add CircleCI and codecov badges

AndreaCrotti:


AndreaCrotti:
Ah I see, well for me it's more about showing the state of a project, and even just showing that you care about tests and test coverage for example.
So yeah I think there is no harm in having them for sure.

PR #287, opened 2020-02-26
Add model validation to the UI

### LINKED ISSUE #272: Add model validation to the UI
Extracted from and blocked by #232.

### UI

I’m thinking a user invokes `fc4` like so:

```man
$ fc4 --validate --model PATH
```

Explained:

* `--validate` is a “feature” flag that activates a new [“feature option”](https://fundingcircle.github.io/fc4-framework/tool/#feature-options) that validates the model
  * Down the road, this would also validate any views that were specified — or maybe even recursively find views and validate them, similar to how the other feature options work
  * If we retain support for Structurizr Express diagrams (see _Open Questions_ in #229) then it would make sense for this feature to support those as well.
* `--model PATH` will be how the user points the tool to the location of the model
  * This would be sort-of optional:
    * If omitted, and a feature option is specified that requires a model, then:
      1. The tool would look for a directory named `model` within the current working directory.
      1. If found, it would use that.
      1. If not found, then the tool would exit with an error message.

### UX

Assuming the tool was invoked successfully/correctly (see UI above) then the UX would be like so:

1. The tool would:
   1. Find all the YAML files in the model
   1. Validate each file against the model schema, keeping track of any problems
1. If all files are valid, it would exit with code 0 and write nothing to stdout nor stderr
1. If any files are invalid, it would:
    1. write a report to stderr
    1. exit with a non-zero exit code (see _exit codes_ below)

### Exit Codes

* When all files in the model are valid, then the exit code should be 0
* When one or more files are invalid, then the exit code should definitely be non-zero, but it’s not obvious what it should be
  * IIRC, @miketheman pointed out to me that it’d be useful for automated tooling, scripts, CI jobs, etc to be able to differentiate between a non-zero error code indicating that the tool failed to operate properly versus one indicating that the model failed validation.
    * Therefore we should probably choose a non-zero exit code that is safe to use, and is probably not -1 or 1, nor any code that’s reserved by operating systems, POSIX specs, etc, and designate that code as having those semantics for this tool.
    * i.e. TODO.

aviflax:
As per #272.

## Notes to reviewers

* Thank you!

----

Closes #272

### DIFF
--- a/src/fc4/dsl/model.clj
+++ b/src/fc4/dsl/model.clj
@@ -246,7 +246,7 @@
     (::anom/message parsed)
 
     :else
-    (expound-str ::file parsed)))
+    (expound-str ::file parsed {:print-specs? false})))
 
 (s/def ::err-msg string?)
 
--- a/src/fc4/io/cli/main.clj
+++ b/src/fc4/io/cli/main.clj
@@ -1,44 +1,50 @@
 (ns fc4.io.cli.main
   "The CLI command that is the primary interface of the tool."
   (:gen-class)
-  (:require [clj-yaml.core :refer [parse-string]]
-            [clojure.pprint :refer [pprint]]
-            [clojure.set :refer [subset?]]
-            [clojure.string :as str :refer [join lower-case split trim]]
+  (:require [clojure.pprint :refer [pprint]]
+            [clojure.set :refer [intersection]]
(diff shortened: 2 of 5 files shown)


aviflax:
@alex-steinbaugh @AndreaCrotti any chance _one_ of you might be able to review this within the next day or two?

PR #289, opened 2020-03-12
Add nightly CI workflow to check deps

aviflax:
Informational only for now but I plan to make it fail when outdated deps are found once Olical/depot#39 is resolved.


mattsoutherden:
Are you planning to just manually inspect the output?

aviflax:
For now, yes, I suppose. There’s not a ton of value in that; I’m thinking the larger value is in establishing a nightly workflow that checks things.

That said, I have done most of the work already to the library `depot` to enable it to fail the job if any deps are out of date. Perhaps I should just finish that and incorporate it before shipping this.

Issue #268, opened 2020-01-13
Structurizr Express is being discontinued


simonbrowndotje:
Hi folks, for a number of reasons, we've made the decision to discontinue Structurizr Express, and it will be unavailable from 1st July 2020. The workspace version of Structurizr (free and paid plans, cloud service and on-premises installation) remains as-is.

I can see from commentary on the proposals (e.g. [EP01](https://github.com/FundingCircle/fc4-framework/blob/7984a94/proposals/ep01-decouple-models-and-views-new-dsl/ep01-decouple-models-and-views-new-dsl.md)) and issues (e.g. #229) that you're considering how to render diagrams specified with the new DSL, and that your new workflow will be much more about forward generating diagrams from your DSL, rather than round-tripping Express definitions. If you're not going to build your own rendering engine, you can still use Structurizr for rendering diagrams with a slightly modified workflow:

 1. Create and upload a Structurizr workspace JSON file via the [Structurizr web API](https://structurizr.com/help/web-api) ... the simplest way to do this is via one of the [Java/.NET/PHP/TypeScript client libraries](https://structurizr.com/help/client-libraries). The [Workspace export and import page](https://structurizr.com/help/workspace-export-and-import) shows a code example.
 2. Open the diagram editor as an _authenticated user_ to edit the diagram layout. Via the [scripting support](https://structurizr.com/help/scripting-diagrams) I've added a way to register a JavaScript callback (the ```onWorkspaceSaved(callback)``` function) that will allow you to write some code that will be executed when the workspace is saved. Using this in conjunction with Puppeteer, it's straightforward to open the diagram editor interactively, and export the PNG/SVG diagrams whenever the workspace is saved. [Here is a working example script](https://github.com/structurizr/puppeteer/blob/master/edit-and-export-private-diagrams.js).
 3. After modifying the layout of the diagram(s) contained in the workspace, the paper size and element x,y coordinates can be obtained by downloading the workspace JSON definition via the web API.
 
If you have privacy concerns with uploading information to the cloud service, you can install the free version of the on-premises installation, which works in the same way. You may see an overall performance improvement too by having a local installation.

[...]

PR #293, opened 2020-06-30
Switch to local instance of Structurizr/Express

### LINKED ISSUE #288: Use the on-premises (local) version of Structurizr Express
Because the Web-accessible version has been discontinued, as per in #268.

For more on the specific reasons to go with this approach, please see [this comment on #268](https://github.com/FundingCircle/fc4-framework/issues/268#issuecomment-592165632) and [this comment on #284](https://github.com/FundingCircle/fc4-framework/issues/284#issuecomment-641446349).

aviflax:
To account for the retirement of the publicly-accessible instance of Structurizr Express, and the concomitant need to run the on-premises version of Structurizr in order to use Structurizr Express to render diagrams.

See FundingCircle/fc4-framework#268

----

Resolves #288


aviflax:
@mattsoutherden this is ready for final review.

test/Dockerfile
```diff
@@ -8,21 +8,19 @@
...
-# create it, it gets created via root with the wrong permissions and then the user that’s used for
-# all the RUN commands (circleci) can’t write to it. This is *really* annoying.
+WORKDIR /home/circleci/project
+
 USER root
-RUN mkdir -p /home/circleci/project && chown -R circleci:circleci /home/circleci
 
-# As long as we’re root, let’s install rlwrap. It’s needed to start an interactive repl via `clj`
+# Work around some bizarre permissions issue.
+RUN chown -R circleci:circleci /home/circleci
```
mattsoutherden, line 16: Does this workaround need to be done here on build and on execution (as in the bin files)?

bin/tests
```diff
@@ -1,15 +1,24 @@
...
+# The max heap size is set to 4GB because I’ve seen the process just die at 3GB and below.
 #
 # We’re including the dependencies of test/coverage here, even though we’re not
 # actually measuring coverage in this script, because we want the computed set
 # of deps to be exactly the same as that computed by ./download-test-deps, which
 # includes the :test/coverage profile, so that a single script can download the
 # deps for both this script and ./tests-with-coverage
-clojure -J-Xmx2g -A:test:test/run -R:test/coverage
+
+docker-compose run fc4 clojure -J-Xmx4g -A:test:test/run -R:test/coverage
```
mattsoutherden, line 22: What's the difference between this and `test:test/coverage` below? (It's just a bit confusing to me)

test/Dockerfile
aviflax: Ah, good catch. Probably not. Normally I’d clean this up, but I’m not in the mood right now. Thank you though!

bin/tests
aviflax: `test` and `test/coverage` are two discrete aliases in the `deps.edn` file. I’m using a convention here: if an alias has a `/` in its name, then the first segment is the name of an alias that is required for the alias to work. So the alias `test/coverage` builds on `test` and will only work when `test` is _also_ included.

Here in *this* invocation of `clojure`, we’re using `-A` to tell Clojure to use all elements of the aliases `test` and `test/run`. We then also use `-R` to tell Clojure to use only the elements of `test/coverage` that are related to building the Java classpath. This is just so that the computed set of dependencies will be the same across invocations, so that the superset of dependencies will be downloaded into the image on build.

In the script `tests-with-coverage`, we include `test/tests-with-coverage` in the list with to `-A` so as to specify that we want to use *all* the elements of that alias, which includes the specification of a “main class” — essentially, an entrypoint — which will run the tests. Therefore we don’t need `test/run`.

(I see now that these aliases should have been named `test/run-tests` and `test/run-tests-with-coverage`.)

For reference, here’s some of the output of `clojure --help`:

```text
$ clojure --help
Version: 1.10.1.536

Usage: clojure [dep-opt*] [--] [init-opt*] [main-opt] [arg*]
       clj     [dep-opt*] [--] [init-opt*] [main-opt] [arg*]

The clojure script is a runner for Clojure. clj is a wrapper
for interactive repl use. These scripts ultimately construct and
invoke a command-line of the form:

java [java-opt*] -cp classpath clojure.main [init-opt*] [main-opt] [arg*]

The dep-opts are used to build the java-opts and classpath:
 -Jopt          Pass opt through in java_opts, ex: -J-Xmx512m
 -Oalias...     Concatenated jvm option aliases, ex: -O:mem
 -Ralias...     Concatenated resolve-deps aliases, ex: -R:bench:1.9
 -Calias...     Concatenated make-classpath aliases, ex: -C:dev
 -Malias...     Concatenated main option aliases, ex: -M:test
 -Aalias...     Concatenated aliases of any kind, ex: -A:dev:mem
```

HTH!

